PR #83, opened 2014-10-25
adding Logitech Media Server strategy

jellea:
Don't know how useful this is because the ip is probably different in other cases.


foozee:
maybe a good option could be to have the address of the logitech server configurable in preferences? that would make it very flexible to users to set the IP themselves ?

Stillness-2:
May be better another way for identification of the Tab with Logitach Media Server?

Stillness-2:
For example something on the page

foozee:
@Stillness-2 I've just had this discussion with @campincarl on #146 .... running JS on all open tabs (each time the user clicks the beardedspice menu icon) seems to me like a not-the-best idea from a performance perspective, no matter how small footprint this JS has (many users, including myself have more than 40+ tabs open at a time) this would kill performance altogether! @campincarl is currently checking the tab title, but basically I still think that allowing configurable URLs for user-hosted services will just save us most of this hassle :)

CampinCarl:
Using the tab title has worked out great for me so far. I've been testing Subsonic this way and haven't had any issues. In fact, BeardedSpice seems to intelligently maintain the correct tab even if I open additional tabs with the same title.

foozee:
I really didn't like logitech media server :( (takes a very long while to pause/skip/play when using the web interface buttons, and even media controls are not consistent across their different themes) however, seemed to be working as far as BS is concerned with the `Default` web interface :) 
also, I've updated it to work with tab title instead of a hardcoded URL, thanks everybody.

PR #98, opened 2014-12-16
Fix whitespace lines

vitorgalvao:
This introduces both manual and automated changes. They all have to do with making whitespace consistent, so they don’t change any of the functionality.
- All BeardedSpice/*Strategy* files have one and only one line before the last `@end`. This means some had it added, while others had them reduced to one.
- All trailing whitespace was removed from all files. This is both from otherwise empty lines (lines that have nothing but spaces) and lines that have code but end with spaces.
- All tabs were replaced with spaces. One or the other should be used. There were even files that mixed both, sometimes even on the same line. One tab was replaced with seven spaces. The number was chosen since it was what seemed to preserve the structure better, according to github’s visualisation.


jayrhynas:
Just a heads-up, this also replaced all "trailing whitespace" and "tabs" in [BeardedSpiceUserDefaults.plist](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/BeardedSpiceUserDefaults.plist) which is actually a binary file, so it could no longer be opened. I've fixed it in #131.

vitorgalvao:
@jayrhynas Apologies. Thank you for fixing it.

jayrhynas:
@vitorgalvao No worries :smile:

PR #100, opened 2015-01-04
Activate when opening menu

nickvdp:
Allows BeardedSpice to reassert control over the media keys without restarting or opening the preferences.

Makes BeardedSpice the active application when the menu is opened, so that SPMediaKeyTap intercepts the key taps instead of deferring to the next application.


nickvdp:
I just made another change that does the same thing in a different way. Now instead of manipulating the active application (which caused a flicker in the current active window) it goes directly to SPMediaKeyTap.

PR #103, opened 2015-01-06
Add Saavn Strategy

Budgetperson:
Add a Saavn strategy for http://saavn.com

### DIFF
--- a/BeardedSpice/SaavnStrategy.m
+++ b/BeardedSpice/SaavnStrategy.m
@@ -26,22 +26,22 @@
 
 -(NSString *) toggle
 {
-    return @"(function(){var e=document.querySelectorAll('#play')[0];var t=document.querySelectorAll('#pause')[0];if(t.className.indexOf('hide')===-1){t.click()}else{e.click()}})()";
+    return @"(function(){var e=document.getElementById('play');var t=document.getElementById('pause');if(t.className.indexOf('hide')===-1){t.click();}else{e.click();}})()";
 }
 
 -(NSString *) previous
 {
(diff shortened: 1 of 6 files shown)


BeardedSpice/SaavnStrategy.m
```diff
@@ -0,0 +1,52 @@
...
+}
+
+-(NSString *) toggle
+{
+    return @"(function(){var e=document.querySelectorAll('#play')[0];var t=document.querySelectorAll('#pause')[0];if(t.className.indexOf('hide')===-1){t.click()}else{e.click()}})()";
+}
+
+-(NSString *) previous
+{
+    return @"(function(){var e=document.querySelectorAll('#rew')[0];e.click()})()";
```
vitorgalvao: Why create a variable at all? That’s unnecessary work: it could just be `document.querySelectorAll('#rew')[0].click()`. But even that would be highly inefficient here. If you’re going to select the first element anyway, you don’t need `querySelectorAll`, you can simply use `querySelector` (i.e. `document.querySelector('#rew').click()`), but you’re clicking IDs, and there can be only one per page anyway, so the best way to go about it would be just using `getElementById`: `document.getElementById('rew').click()`.

Budgetperson:
Okay I don't think this fixed the whitespace. I'll try again, make sure it works, commit and squash.

vitorgalvao:
The changes are still shown, so it seems like nothing changed.

foozee:
@Budgetperson  we're trying to get that project running again and would really like to include ur PR in the release intended for 14th of April, Would be really great if you could help get that PR in the release, would you? :)

PR #104, opened 2015-01-06
Small changes to Pandora’s strategy

vitorgalvao:
These are mainly cosmetic, and don’t change functionality. Using `querySelectorAll` to select the first (`0`) element is unnecessary: we can simply use `querySelector`. In addition, `next` and `pause` have no reason to be functions.


josefalcon:
Thanks!

PR #118, opened 2015-02-09
Added non-Apple keyboard Previous/Next support

dublinus:
receivedMediaKeyEvent checks for next and previous keys in addition to
fast and rewind. Tested on WASD keyboard and Apple wireless keyboard.


maxdjohnson:
works for me, thanks dublinus

dublinus:
No problem, I'm glad it helped someone.

Issue #108, opened 2015-01-15
Pause/play button stops working after a while


radex:
When I first connect a tab, media keys work perfectly fine. But after a while, iTunes hijacks them, and they stop working with BeardedSpice. If I try to reconnect, it still doesn't work. I have to quit iTunes, or quit and open BeardedSpice again for it to work again.

Am I the only one with this issue?

Duros:
I can confirm this while using safari 8.0.2.

acadet:
I have the same one, using Chrome 40.0.2214.111.
Any news about this?

jayrhynas:
You can re-associate them with beardedspice by activating the app. Unfortunately just opening the menubar dropdown isn't enough, you have to open the preferences window.
iTunes shouldn't randomly hijack focus, but how it works is the media keys will be associated with the most recently activated media app, so if you switch to iTunes at all it will grab the media keys.

Issue #29, opened 2014-02-13
BeardedSpice no longer "works" after having opened another mediakeys-activated application


tobyhinloopen:
I have some issues where BeardedSpice stops recognising mediakeys "randomly". After some amount of testing and figuring out, I finally found a way to reproduce the issue. Here it is:

- Start Chrome

Note that I already had some tabs opened with some unrelated websites that don't have sound playing capabilities. I added a new browser tab for my audio source.

- Start playing a YT video and have it remain active
- Start BeardedSpice

BeardedSpice knows which video is playing correctly, but media keys don't work.

- With the YT video tab active, press the "Set Active Tab" shortcut

Media keys now work properly.

- Start VLC Player or iTunes

Media keys are now assigned to VLC or iTunes

- Quit VLC Player or iTunes

BeardedSpice still knows which video is playing, but media keys are ignored and iTunes does not start either when Play is pressed. Additionally, using the "Set Active Tab" shortcut on the YT video tab doesn't make any difference: Media keys are still ignored completely.

- Quit BeardedSpice

Mediakeys still don't start iTunes.

- Start BeardedSpice

BeardedSpice knows which video is playing correctly, but media keys don't work.

- With the YT video tab active, press the "Set Active Tab" shortcut

BeardedSpice is now working properly again.

I find there are multiple things wrong here. I think BeardedSpice should work "instantly" with the first (or any) applicable browser tab without having to "activate" it first. Additionally, I don't get why iTunes doesn't start with the "play" key anymore after I closed BeardedSpice.

Issue #20, opened 2014-01-28
Stops working after sleep


aristidesfl:
It stops recognising the media keys after the computer has slept and woke up.

Mavericks OS X 10.9.1
BeardedSpice 0.1.0

foozee:
Solved :)

aristidesfl:
:+1:

PR #122, opened 2015-02-18
Aggressively restore media keys subscription

SpeCT:
Hey guys. This patch solves #20, #29 and #108 for me.
PS: I'm not a Cocoa guru, so don't kick me hard ;)


1ps0:
@depressiveRobot I think this patch solved the problem at the time. Can you open a new ticket and be as specific about the problem as possible? App and OS versions are a good start, and browser, behavior, and reproduction steps also help with bugs like this.

PR #126, opened 2015-03-06
Add support for Netflix

martijnengler:
Regarding issue #61 : support for Netflix. I wanted it too, so I built a simple Strategy for Netflix.com


fmborghino:
This is closed by #290

PR #133, opened 2015-03-18
Add favorite function for SoundCloud strategy

frdmn:


frdmn:
@Stillness-2 I noticed that its not the same behaviour as the `L` key like short cut. If you listen a track out of a playlist the `l` key actually likes the track and not the playlist like the solution right now. I tried to implement something like keyPress, but it didn't work out for me so I've used this alternative one.

foozee:
ummm, @frdmn in fact, I've tested your commit before merging it ... have to admit that testing didn't include the case of "liking a track on a  playlist" now I verified it and could reproduce the issue, you're right this won't work :( (because it assumes that the user is on the stream page to work)

I think that using the player like button is a better solution in fact, because we try to use page elements that will be always found on any page no matter where the user is on the site :) 

another thing is, I noticed that no matter where you press `L` on soundcloud, it will always like the currently playing track, not the playlist of which that track is, and also for consistency in BS the user should expect the favorite shortcut to `favorite` the currently playing track not the playlist.

I'll fix that later tonight because I have to go out now  :)

foozee:
fixed by 2b8d33cc0101c1b5bb65f267eaba0993b7e8c17a

PR #136, opened 2015-03-30
Add Support for noadradio.com

foozee:
Hi,

I'm more of a web guy, so I know JS way better than I know any objective-c :D but u guys made it pretty easy to add a new strategy for a site without actually knowing so much objective-c :) 

please review the pull request and let me know if I need to enhance/update anything there.

Kind Regards,
Mahmoud


foozee:
I'm not particularly sure why this build is failing, a previous one missed the podfile.loc and the .xcodeproject files, so I added them but now the build is timing out.

I tried running release.sh and it seemed to work fine and ended outputting "build succeeded", would appreciate any guidance with that, thanks :)

foozee:
I think that the currently failing build needs to be just re-run, because I've tried travis with the updated commit at https://travis-ci.org/foozee/beardedspice/builds/56541025 and it worked :) thanks!

PR #151, opened 2015-04-24
thedrop.club support

domjitsu:


foozee:
Thanks for your PR @domjitsu would be amazing if you could also implement the `trachInfo` method, update the README.md, and undo changes to Podfile.lock if you don't have time though, I can still do it myself when I come to merge this PR :)

PR #157, opened 2015-05-08
Add support for NoonPacific.com

benregn:


Stillness-2:
Everything is fine, but... Please, we need implementation of isPlaying method! :)

case:
Noon Pacific is :dancer: :dancer: so good. Looking forward to seeing this merged + shipped!

PR #175, opened 2015-06-16
Default notification

jayrhynas:
If a notification is requested but the current strategy doesn't provide track info, nothing is shown. This is fine when controlling the player, but unexpected (for the user) when explicitly pressing the Track Info hotkey.

This PR adds a default notification to be used as a fallback when the app is unable to get track info (unsupported by current strategy, or nothing playing in the current native app). It only gets triggered on the notification hotkey callback - all other keys continue to show no notification.

bs-notification


jayrhynas:
I'd like to get others' opinions on whether this is a feature people want before I merge this in.

foozee:
that's kinda cool :)

PR #181, opened 2015-06-25
Add Emby Strategy

ayyjayess:
I have added a strategy for [Emby](http://emby.media/) (previously known as MediaBrowser). 

Emby runs on the localhost and so I am checking both localhost and 127.0.0.1.

### DIFF
--- a/BeardedSpice/MediaStrategies/EmbyStrategy.m
+++ b/BeardedSpice/MediaStrategies/EmbyStrategy.m
@@ -57,14 +57,13 @@
 
 -(Track *) trackInfo:(TabAdapter *)tab
 {
-    NSDictionary *currentItem = [tab executeJavascript:@"MediaPlayer.currentItem"];
-    NSString *imageUrl = [tab executeJavascript:@"document.querySelector('.nowPlayingImage > img:first-of-type').src"];
+    NSDictionary *metadata = [tab executeJavascript:@"(function(){ var ret = MediaPlayer.currentItem; ret['imageUrl'] = document.querySelector('.nowPlayingImage > img:first-of-type').src; return ret;})()"];
     
     Track *track = [[Track alloc] init];
-    track.track = [currentItem objectForKey:@"Name"];
-    track.album = [currentItem objectForKey:@"Album"];
-    track.artist = [currentItem objectForKey:@"AlbumArtist"];
-    track.image = [self imageByUrlString:imageUrl];
+    track.track = [metadata objectForKey:@"Name"];
+    track.album = [metadata objectForKey:@"Album"];
+    track.artist = [metadata objectForKey:@"AlbumArtist"];
+    track.image = [self imageByUrlString:[metadata objectForKey:@"imageUrl"]];
 
     return track;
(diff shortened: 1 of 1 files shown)


andjscott:
Emby can also be run from another computer on your local network using that other computers local IP. Would it be acceptable for the predicate to be based solely on the port in use? i.e. ":9086" I think this is the first locally run webapp on BeardedSpice so I don't know the guidelines.

Stillness-2:
see subsonic strategy

BeardedSpice/MediaStrategies/EmbyStrategy.m
```diff
@@ -0,0 +1,72 @@
...
+
+-(NSString *) displayName
+{
+    return @"Emby";
+}
+
+-(Track *) trackInfo:(TabAdapter *)tab
+{
+    NSDictionary *currentItem = [tab executeJavascript:@"MediaPlayer.currentItem"];
+    NSString *imageUrl = [tab executeJavascript:@"document.querySelector('.nowPlayingImage > img:first-of-type').src"];
```
Stillness-2: Will be better if perform one call of the executeJavascript method for request data.

andjscott:
Emby doesn't keep 'Emby' in the title. The title is just the current playing album name

Stillness-2:
@andjscott This is problem :) Because using executeJavascript for determine of the Emby tab is not acceptable. :) You may connect with Emby developers and suggest them to add "Emby - " as prefix in all page titles, for example, or to change urls from http://localhost:8096/web/xxx to http://localhost:8096/super-emby/web/xxx. :)

Stillness-2:
@moonarch I changed my mind about the "executeJavascript" in the method "accepts". :) Needs try and to see performance.

RSully:
Sounds similar to the Plex Web strategy I wrote. I had originally used the [URL](https://github.com/beardedspice/beardedspice/commit/186148773f4cf1975f80458133b182b874a68ad6#diff-c8004b87f3f206aa724d46b58379bf6fR17) to match but that was changed to [execute JS](https://github.com/beardedspice/beardedspice/commit/1ab165fd3388f0178828cc20733b3a2212264f73#diff-c8004b87f3f206aa724d46b58379bf6fR16).

ianawilson:
This looks like it wouldn't support the case where the Emby server is running on another computer, which is a common use case, I believe, as well as my own use case. :) Is switching away from URL/port and moving to entirely `executeJavascript()` (like @RSully's Plex Web strategy) an acceptable strategy?

I'm definitely interested to see this PR finished and accepted, if that's a possibility.

nVitius:
@1ps0 This request is really old and significantly out of date. Should we close it? If anyone else cares for an emby strategy, they can always create an issue for it.

1ps0:
@nVitius lol getting tunnel vision here. I'll reevaluate it properly once i'm no longer in transit. Thanks :)

1ps0:
@ayyjayess Thanks for the PR :) However, I'm closing this pending reevaluation/demand.

PR #223, opened 2015-08-20
Add initial Plex Web support.

### LINKED ISSUE #56: Please add Plex support
app support

RSully:
This adds initial support for Plex. I reached out to see if there is a better way to handle this than just clicking DOM elements, as that could break at any point, but it works.

This fixes #56.

Known issues:

I am experiencing problems when I have 2 Plex Web windows open at
the same time. However, in normal usage I suspect this is more of an
edge-case than anything else.

(I am not convinced this isn't just an issue with BeardedSpice in general anyways)


sickill:
:heart:

Stillness-2:
Anybody will test please Plex strategy, which is in master branch, that it works with remote plex server, i.e. through http://app.plex.tv/web/app .

matrixagent:
I'd be happy to test it, didn't succeed in building it myself though.

RSully:
Compiled fine here on 10.10 Xcode 7.0.1 Cocoapods 0.38.2.

Confirmed works for music and videos using local server (192.168.x.y:32000/web). I don't use plex.tv.

ascagnel:
A build against master works with plex.tv for videos and music.

Stillness-2:
Thanks! :)

ascagnel:
Also, it works with the new pinned tab functionality in Safari.

PR #241, opened 2015-10-02
Add Superplayer media strategy

thiagodorneles:
Adding media strategy for Superplayer website.
Implemented the functions to the play, pause, forward and to love music which is playing.


Stillness-2:
Not fully implemented

Issue #208, opened 2015-07-30
[App Support] Microsoft's Groove (formerly Xbox Music)


samtwil:
I use this app all the time for YouTube playback, thank you so much for all the hard work!

Could we get support for the new Groove music player at music.microsoft.com?

enjoyfebruary:
Second. I don't believe much has changed from when Groove was Xbox Music except for the URL. Hopefully this can get taken care of without much effort.

PR #250, opened 2015-10-12
Fix the XboxMusic strategy after its URL change

lypanov:
This fixes the issue reported at:

https://github.com/beardedspice/beardedspice/issues/208


1ps0:
Doing rounds of updates. Please verify this change is still valid. Thanks :)

PR #265, opened 2015-11-02
Added support for Webkit nightly builds

JackoPlane:


Stillness-2:
Have bug with pinned tabs.

PR #280, opened 2015-11-20
Closes #279 - Rhapsody Support

### LINKED ISSUE #279: Rhapsody Support
Would love support for the Rhapsody web player

lolpack:
This PR adds toggle, pause, next, like and previous support for the Rhapsody web player. The friendly album name isn't always available without an extra API call. If it's not available, it should return nothing. 

Please let me know if there's anything I forgot and I will update :)

@Stillness-2


Stillness-2:
We need  `isPlaying` method

lolpack:
@Stillness-2 added `isPlaying` method

Issue #293, opened 2015-12-07
Overcast playback control not working


loganrockmore:
Overcast playback control has recently stopped working for me. The tab within Google Chrome shows up in the BeardedSpice menu, but selecting it will not check it as the active playback source and so no media key functionality ends up working. I'm using BeardedSpice v1.2.2 on OS X El Capitan v10.11.1.

nelox:
I can confirm the same issue using:
* Beardedspice 1.2.2
* Chrome 47.0.2526.80 (64-bit)
* OS X Yosemite 10.10.5 (14F1505)

fmborghino:
@Stillness-2 I think maybe whatever broke Chrome 46 has been reverted. I was able to get this working with Chrome 47 by limiting that workaround to only v 46. Just throwing this out there for those that understand that fix better. Tentative change:

```
diff --git a/BeardedSpice/Tabs/ChromeTabAdapter.m b/BeardedSpice/Tabs/ChromeTabAdapter.m
index 6fb03aa..1b11973 100644
--- a/BeardedSpice/Tabs/ChromeTabAdapter.m
+++ b/BeardedSpice/Tabs/ChromeTabAdapter.m
@@ -29,7 +29,7 @@
     out.applescriptIsolatedVersion = NO;
     NSInteger version = [[(ChromeApplication *)application.sbApplication version] integerValue];
     NSArray *googleBundleIds = @[APPID_CANARY, APPID_CHROME, APPID_CHROMIUM];
-    if (version > 45 && [googleBundleIds containsObject:application.bundleIdentifier]){
+    if (version == 46 && [googleBundleIds containsObject:application.bundleIdentifier]){
         out.applescriptIsolatedVersion = YES;
     }
```

Stillness-2:
@fmborghino No, this is another problem :) 
On my environment all works fine:
BS 1.2.2 (14)
GH 47.0.2526.80 (64-bit)
OSX 10.11.1 (15B42)

fmborghino:
Ah ok! Meantime I updated the missing API bits in the strategy #295 ... not going to merge that just yet in case there's some impact on this issue.

PR #295, opened 2015-12-09
Update Overcast strategy for new API

fmborghino:
- added isPlaying, pause
- added full track info


fmborghino:
Merging this as it should be orthogonal to #293 and is good for those who don't have that problem.

PR #296, opened 2015-12-09
Fix UdemyStrategy and add symlink to fix broken CocoaPods

Coder-256:
Newer versions of Chrome break this strategy because the strategy no
longer has access to jQuery. Updated to be independent of jQuery. Xcode
seemed to want Pods.xcodeproj to be in the main project folder, so I
added a symlink.


Stillness-2:
A little bit wrong :)

Stillness-2:
Thanks for the tip :)

Coder-256:
That works. But this version is more optimized and doesn't depend on jQuery, so you should consider still merging it.

Stillness-2:
"Xcode
seemed to want Pods.xcodeproj to be in the main project folder, so I
added a symlink." - I don't understand :) What version of Xcode are you use?

Coder-256:
I am using Xcode 7.1

PR #302, opened 2015-12-18
Added dynamic strategy updateability

1ps0:
Added the ability to check for new strategy updates without downloading a new version of the binary.

In essence, I've distilled all the various *Strategy.[mh] files into their core Javascript, storing all the relevant data in plists. The strategy plists can be found in the same folder as the original Strategies.

#### How It Works
The approach is simple. We lock the source for all updated javascripts to the master beardedspice repository. When a new PR is accepted and imported, it immediately becomes available for all users running the updatable version.

An index file containing the names of all available strategies and their up-to-date version numbers is stored in the MediaStrategies folder: [versions.plist](https://github.com/1ps0/beardedspice/blob/versioned-strategy-plists/BeardedSpice/MediaStrategies/versions.plist).

The user can enable the option to Check for Updates on Login (disabled by default), or just runs Check for Updates in the main menu. When updates are checked, the app downloads the above-mentioned versions.plist file, and checks each entry to see if it's new or the version number has been increased. 

When a new/updated strategy has been found, it downloads it to the ~/Library/Application Support/BeardedSpice/MediaStrategies/ folder. Then when the strategy is used, the cached copy is refreshed from file and loaded in its stead.

#### Tests

Update: ended up removing these in order for TravisCI to pass (it was getting hung up somewhere). This is fine, as it'll give me a chance to better put together the package I intended.


Stillness-2:
No offense, but... You did such a great job, but why? What is the point of this idea? What does it do? What are the benefits of this idea compared to the approach which is used now? :)

1ps0:
It's a solution to a problem that emerges at scale. Currently the project is just shy of 60 (?) strategies. All of them homogenized js patterns. 

First, this simplify coding for new strategies. Second, it helps avoid the pitfall thats intrinsic to the project, namely playing catchup with a large number of streaming websites. 

Take a look at the number of pending PRs for adding or fixing different sites. With the current plan, every new PR that's integrated would require users to download and 'install' a new binary. With the new plan, they can have have it automatically.

Does that make sense? I've bounced this back and forth a lot, and it seemed the right time to push it.

foozee:
The idea is that with such a technique, we can allow users to (just to name a few of possibilities this opens):
* Download or remove media strategies as if they were plugins
* Update individual media strategies without having to update BS binaries itself
* Write and test new media strategies without any knowledge about Objective-C
* ....

Stillness-2:
In this form, as you suggest, this does not solve the main problem. Maintainers  not have the desire to test new strategies. :) In the past, more than half of the strategies that come in the form of PRs contain logic (not syntax) errors. This leads to the conclusion that you can't take and simple to merge. To check you need to register on these sites, sometimes give data of payment cards, sometimes to pay for a subscription. 

Again, the main problem is that maintainers don't maintain.

The ability to write only on javascript has no benefits in itself, in my opinion.
And updating of strategies, also does not solve the main problem. 

This idea makes sense, if carried to their logical conclusion. 
Implement the interface for creating/editing strategies in the application with ability of import/export strategies. Then the exchange of strategies will be possible without a project, so it will take away the responsibility for the wrong functioning strategies that are not available in the distributive.

I suggest you do more work (Plugins, UI.. ) :)

foozee:
I've earlier started working on a chrome plugin that tries to make creating a new media strategy a matter of a couple clicks, I could pass some extra time on that to see if it makes any sense then :)

p.s.: no one said maintainers don't maintain :) I guess this is more of an effort to take off some of the effort off our shoulders :)

1ps0:
I'd consider the improvement in maintenance to be a side effect. 

The primary goal is to improve the user experience. Unless you're constantly up to date with the latest version (which I guarantee most wont), their ability to control their music will become increasingly out of date.

Everything else is biscuits and gravy.

1ps0:
Thoughts?

foozee:
:+1:

Stillness-2:
I already said my offer. :) 

Encourage other participants to speak "yes" or "no" for PR, in the form as it exists now, and will decide majority of votes. :)

I say "no". :)

But if others participants will accept your variant. There are some issues:
1. Menu item "Check for Updates" can't have this name, because in "distr" branch such item exists already. May be "Check for Strategies Updates"
2. We need in UI when user presses this item, like a progress indicator and output of the checking results, in modal dialog, for example.
3. In preferences, where list of the strategies, we show "A" indicator for strategies, only where "isPlaying" method implemented.

foozee:
:+1: @Stillness-2  actualIy I haven't checked the PR itself, I'm just agreeing with the idea/concept itself :)

1ps0:
@Stillness-2 Appreciate the feedback, I'll definitely take a look and update the PR shortly. 

I missed the core thoughts behind your initial response. I've considered a longer term approach to this feature, which includes some of what you proposed, but the focus of this specific PR is an MVP (minimum viable product) conversion of strategies. Already the branch is out of date vs recently accepted code. My current target is to get this part live so we don't keep bashing heads. 

The tests I had initially were targeted at testing a cached copy of a website against the accepted javascript, where the javascript contained 100% of the page-specific logic. Considering I was able to convert all 65 of the sites to use this method without any removal of functionality, I'd say it is a very feasible approach for the future.

1ps0:
Updated to address concerns that were raised. Any other feedback or comments appreciated.

Coder-256:
**+1** :+1: I say a HUGE YES!!! This would make it a lot easier to code and require no knowledge of Objective C, and more as detailed in #332. This is a great solution for part 1 of the second comment on that issue.

1ps0:
Coming back to this after vacation with fresh eyes. @Coder-256 Thank you for the endorsement.

First, I'm still willing to maintain and update this PR and related code. I still see the feature as necessary for future growth.

I will reiterate my points, because they are important, and seem largely missed.

1. **All site-specific code is in javascript.** _No custom logic needs to be implemented in Objective-C_. I've already demonstrated this in NoAdRadioStrategy under `trackInfo:` with the use of `usleep()`. It has been tested and shown to work. @Stillness-2 

2. All strategies are **secure**. This is done by keeping updates limited to the master repository. Only accepted PRs will be deployed to users.

3. This is a high level module equivalent. @Stillness-2  In your words in #332: "Good idea. Will be interesting if all logic of "players" will be separate from AppDelegate.m into** "one high level module"**."
**I have already done this.**

4. This feature enables **hot deployment** of fixes. As in, if a bad JS snippet is imported, it can be fixed without End of Life versions floating around running bad, or even malicious code.

5. Finally, **this is not the final form**. Once this component is integrated, I intend not only to maintain it, but to extend it with proper tests for JS (as mentioned before) and later features (plugins for example). 

There is always more work that can be done. I ask you look at the work _already_ done.

Stillness-2:
I all already said in a previous posts. :) And I'm not ignoring you. And I don't know why other people (https://github.com/orgs/beardedspice/people) don't speak out and ignore this matter. Here we have "democracy" almost. :))) 

And again: if your option would be accepted by others, please at least follow the modifications described in the previous post. :)

fmborghino:
As a general principle I like the idea of decoupling the “data” of the strategies from the app itself.

As discussed above there are several possible issues this design might address.

I’ll rehash those issues and comment on them and on how this PR touches them, as I see it.

**# 1** Strategy updates without a new binary mean more frequent or on-demand strategy updates.

This is a big bonus and I think we all recognize the general benefits of that here. This change is a good start to a mechanism for that: PR with only js changes is created. Maintainers test (as usual) and merge. Users get the update. So far so good.

**# 2** The main bottleneck with strategy updates is maintainers validating the changes.

Stillness-2 makes this point, and it is valid. The benefit in # 1 is still bottlenecked behind # 2. It’s still quicker (no need to wait for a new binary) but the bigger bottleneck is still there. Why is this part slow? Well, part time job here 😃 and...

Maintainers need to do a good job no matter what:
- Look for malicious PRs.
- Figure out new sites and their features, see if the PR works as advertised.
- Often there are paywalls / language issues / and other roadblocks <- this one has stopped me on a few PRs

The decoupling strategy *could* make this a lot easier (Stillness-2 alluded to this): If there were also an import feature for testing ahead of the PR becoming merged and available for automatic update. This would allow non-maintainers who use the sites to test and give a thumbs-up in the comments. Maintainers could then give the PR a further code-review for sanity check/malicious code etc without needing to fire up the site and test in detail. This would effectively grow the pool of “strategy code reviewers” without needing to grow the maintainers pool.

*This really helps with the scaling problem*

**# 3** Code embedded in code is always a pain in the *** to work with. Having js as strings in Obj-C is no fun. Having it in plists is no better. If we are going to go this route, we could also consider a direction (sure, future) where the strategy can be handled as a plain .js file with all the methods, or a directory with several js files per method. We can talk about the details. Something that is more native for strategy coders in js would make it easier in general.

This is just convenience, but this general strategy of decoupling would allow this at some point.

OK so how about this PR? I think to the extent that it opens the door to improving all 3 of the problems above, I like it. I vote “yes” BUT I’d like to see Stillness-2’s requests baked in - he has the most experience in this code lately (and I’ve not worked on the core app, only on strategies, so I am not in a good position to review the implementation details).

You’ve gone a long way to solve # 1, we might improve # 2 with the ability to import proposed changes, and # 3 is a nice to have.

Stillness-2:
@1ps0 There is such a proposal. Make a separate branch. Bring it in accordance with the latest changes in the master branch. I'll check again. And..

Coder-256:
@1ps0 @fmborghino @foozee

@Stillness-2, Should people who want this feature just +1 this thread?

1ps0:
@Coder-256 I think so, yes.

FYI I'm in the process of updating against the master branch. Life things happening though so it will take a little longer than expected.

1ps0:
@Stillness-2 that refactor yesterday didn't trip me up much. 

In the process of testing the merge, making sure I didn't leave anything out.

1ps0:
Looks solid from where I'm standing. Pull at your leisure @Stillness-2

BeardedSpice/AppDelegate.m
```diff
@@ -388,23 +383,63 @@ - (void)headphoneUnplug{
 #pragma mark Actions
 /////////////////////////////////////////////////////////////////////////
 
+- (IBAction)checkForUpdates:(id)sender
+{
+    NSMenuItem *item = [statusMenu itemAtIndex:0];
```
Stillness-2: bug `itemAtIndex:0 `

BeardedSpice/AppDelegate.m
```diff
@@ -388,23 +383,63 @@ - (void)headphoneUnplug{
...
+    NSString *title = item.title;
+    dispatch_async(dispatch_get_main_queue(), ^{
```
Stillness-2: I think will be better `dispatch_sync`

BeardedSpice/AppDelegate.m
```diff
@@ -388,23 +383,63 @@ - (void)headphoneUnplug{
...
+    });
+
+    __weak typeof(self) wself = self;
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH,0), ^{
+        __strong typeof(wself) sself = wself;
+
+        BSStrategyVersionManager *manager = [BSStrategyVersionManager sharedVersionManager];
+        NSUInteger updateCount = [manager performSyncUpdateCheck];
+
+        NSArray *strategies = [MediaStrategyRegistry getDefaultMediaStrategyNames];
```
Stillness-2: if `updateCount` = 0, why we reload strategy.

BeardedSpice/AppDelegate.m
```diff
@@ -388,23 +383,63 @@ - (void)headphoneUnplug{
...
+        for (NSString *name in strategies)
+        {
+            BSMediaStrategy *strategy = [BSMediaStrategy cacheForStrategyName:name];
+            [strategy reloadData];
+        }
+
+        [sself refreshTabs:nil];
+
+        dispatch_async(dispatch_get_main_queue(), ^{
```
Stillness-2: `dispatch_sync`

Stillness-2:
@1ps0  I created separate branch `versioned-strategy` from that PR, and invite you to work on this.

PR #307, opened 2015-12-31
Add Bugs Strategy

sookcha:
FYI : Bugs is a South Korean music streaming service.


Stillness-2:
You must implement `isPlaying` method too.

Stillness-2:
hm

PR #309, opened 2016-01-04
Fixed all warnings, added "Show" functionality in Notifications

saagarjha:
Created stub methods to satisfy protocols
Fixed a selector
Make Notifications more useful by making the "Show" button in banner alerts actually show the active application
screen shot 2016-01-03 at 21 42 14

### DIFF
--- a/BeardedSpice/AppDelegate.m
+++ b/BeardedSpice/AppDelegate.m
@@ -61,15 +61,15 @@ BOOL accessibilityApiEnabled = NO;
     // Insert code here to initialize your application
     // Register defaults for the whitelist of apps that want to use media keys
     NSMutableDictionary *registeredDefaults = [NSMutableDictionary dictionaryWithObjectsAndKeys:
-                        [SPMediaKeyTap defaultMediaKeyUserBundleIdentifiers], kMediaKeyUsingBundleIdentifiersDefaultsKey,
-                        nil];
-
+                                               [SPMediaKeyTap defaultMediaKeyUserBundleIdentifiers], kMediaKeyUsingBundleIdentifiersDefaultsKey,
+                                               nil];
+    
     NSDictionary *appDefaults = [NSDictionary dictionaryWithContentsOfFile:[[NSBundle mainBundle] pathForResource:@"BeardedSpiceUserDefaults" ofType:@"plist"]];
     if (appDefaults)
         [registeredDefaults addEntriesFromDictionary:appDefaults];
     
     [[NSUserDefaults standardUserDefaults] registerDefaults:registeredDefaults];
-
+    
     // Create serial queue for notification
     // We need queue because track info may contain image,
     // which retrieved from URL, this may cause blocking of the main thread.
@@ -77,13 +77,13 @@ BOOL accessibilityApiEnabled = NO;
     //
     
     [[NSDistributedNotificationCenter defaultCenter] addObserver:self selector:@selector(interfaceThemeChanged:) name:@"AppleInterfaceThemeChangedNotification" object:nil];
-
+    
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(generalPrefChanged:) name: GeneralPreferencesNativeAppChangedNoticiation object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(generalPrefChanged:) name: GeneralPreferencesAutoPauseChangedNoticiation object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(generalPrefChanged:) name: GeneralPreferencesUsingAppleRemoteChangedNoticiation object:nil];
     
     [[NSNotificationCenter defaultCenter] addObserver: self selector:@selector(receivedWillCloseWindow:) name: NSWindowWillCloseNotification object:nil];
-
+    
     [[NSUserNotificationCenter defaultUserNotificationCenter] setDelegate:self];
     
     [self setupPlayControlsShortcutCallbacks];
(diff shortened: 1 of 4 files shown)


BeardedSpice/Preferences/MediaControllerObject.m
```diff
@@ -17,7 +17,7 @@ - (id)initWithObject:(id)object{
         if ([object respondsToSelector:@selector(displayName)]) {
             _name = [object displayName];
         }
-        if ([[object class] instancesRespondToSelector:@selector(isPlaying)] || [[object class] instancesRespondToSelector:@selector(isPlaying:)]) {
+        if ([[object class] instancesRespondToSelector:@selector(isPlaying)] || [[object class] instancesRespondToSelector:@selector(isPlaying)]) {
```
Stillness-2: what

BeardedSpice/MediaStrategy.m
```diff
@@ -68,6 +68,11 @@ -(Track *) trackInfo:(TabAdapter *)tab
     return NULL;
 }
 
+// Method stubs to satisfy the TapAdapter protocol
+-(BOOL) isPlaying:(TabAdapter *)tab {
+    return NO;
+}
+
```
Stillness-2: this is wrong

BeardedSpice/AppDelegate.m
```diff
@@ -151,6 +151,13 @@ - (BOOL)userNotificationCenter:(NSUserNotificationCenter *)center shouldPresentN
     return YES;
 }
 
+// When "Show" clicked on a Notification
+- (void)userNotificationCenter:(NSUserNotificationCenter *)center
+       didActivateNotification:(NSUserNotification *)notification {
+    // Raise the currently active tab
+    [activeTab activateTab];
```
Stillness-2: May be
```
[self autoSelectedTabs];
[activeTab toggleTab];
```

Stillness-2:
We have hotkey, which raises active app and active tab.  So you propose to duplicate the functionality.
Also in your xcode has its own formatting options that are different. Therefore, PR includes changes, where there is no change.

BeardedSpice/Tabs/NativeAppTabAdapter.m
```diff
@@ -27,6 +27,30 @@ + (NSString *)bundleId{
...
+}
+
+- (Track *) trackInfo {
+    return nil;
+}
+
+- (BOOL)isPlaying {
+    return NO;
+}
+
```
Stillness-2: This is wrong

saagarjha:
Ok, let's go through it:
> We have hotkey, which raises active app and active tab.

This is the most important, and the main reason for this PR. I am aware of the hotkey functionality, and I'm not duplicating it. It's just that the Bearded Spice Notification has a button that says "Show", and a user would expect it to "Show" whatever the Notification is about; in this case the currently playing track. I've simply added that functionality; this is *not* a duplicate of current functionality.

> This is wrong

Could you explain this more please? The warnings were bothering me, do you not want the `TabAdapter` protocol's methods implemented? Do you want other default values?

> what

Was resolving warnings, trying to see what Xcode was complaining about. Did not realize I left it like, without fixing it. Is suppressed in 6cbfba0.

> May be
   `[self autoSelectedTabs];`
   `[activeTab toggleTab];`

Seems to have a lot of stuff setting it to the selected tab and whatnot. When the notification is presented and the user responds, the selected tab shouldn't change. Thus, I thought that directly raising the tab should be enough.

> Also in your xcode has its own formatting options that are different.

I have tried to make my formatting match in 03839ed. Please let me know if where I need to change the formatting if necessary.

Stillness-2:
>Could you explain this more please? 

Check Preferences -> Media Controllers table in your case. All rows contain icon 'A', but this wrong. You implemented `isPlaying` in all classes of the media strategies, including old strategies that do not support it.

Stillness-2:
>This is the most important, and the main reason for this PR. I am aware of the hotkey functionality, and I'm not duplicating it. It's just that the Bearded Spice Notification has a button that says "Show", and a user would expect it to "Show" whatever the Notification is about; in this case the currently playing track. I've simply added that functionality; this is not a duplicate of current functionality.

Have an idea to make a action on a notification, that to find the current track in famous search engines :)

Stillness-2:
>Please let me know if where I need to change the formatting if necessary.

If you looking for `AppDelegate.m` in PR, you will see many changes, for example in the method `setupFavoriteShortcutCallback` and so on. But i do not believe that they exists.

Stillness-2:
>Thus, I thought that directly raising the tab should be enough.

No, because you may move tabs into Safari window before this, for example.

saagarjha:
Ok, I've reverted and fixed it based on your suggestions.
The formatting should be back to normal.
I've also removed the stub methods and replaced them with warning suppressions. While I still think this is unsafe (wouldn't calling them cause a crash? How would a client know not to call them?) it appears that certain classes rely upon selectors and their existence was messing them up.
Finally, I've removed the raising functionality. I'm not quite sure what you mean by 

> find the current track in famous search engines

Do you want to open a page on Google with the track name? Search for it on iTunes?

Stillness-2:
> Do you want to open a page on Google with the track name? Search for it on iTunes?

Yes, kind of

Stillness-2:
We need tab in Preferences, in which user may select appropriate search engine for him.

saagarjha:
Sure, I'll work on it when I have time. Do you have any way you want it implemented?

Stillness-2:
Sorry, but i close this PR. I take #pragma commands from here. :)

PR #311, opened 2016-01-05
Add Radio Swiss Jazz Strategy

Geekfish:
This adds support for [Radio Swiss Jazz](http://www.radioswissjazz.ch/).
It should work both on the full page and on the popup version.

I implemented as much of the API as I thought was possible. `next`, `previous`, `pause` and `favourite` cannot be implemented as they are not supported by the media provider.

Please let me know if this requires any improvements :)


Stillness-2:
Getting track info causes app crush
Reason is error, when we get image url:
https://www.dropbox.com/s/ncyszxd9f5kb1cv/%D0%A1%D0%BA%D1%80%D0%B8%D0%BD%D1%88%D0%BE%D1%82%202016-02-14%2014.41.41.png?dl=0

BeardedSpice/MediaStrategies/RadioSwissJazzStrategy.m
```diff
@@ -0,0 +1,71 @@
...
+    NSString *baseImageUrl = @"http://www.radioswissjazz.ch";
+    NSDictionary *metadata = [tab executeJavascript:@"(function(){ return {"
+                              @"  image:  document.querySelector('#live img.cover').getAttribute('src'),"
+                              @"  album:  document.querySelector('#live img.cover').getAttribute('title'),"
+                              @"  track:  document.querySelector('#live .title').innerText,"
+                              @"  artist: document.querySelector('#live .artist').innerText,"
+                              @"}})()"];
+
+    track.track = [metadata valueForKey:@"track"];
+    track.image = [self imageByUrlString:[baseImageUrl stringByAppendingString:[metadata valueForKey:@"image"]]];
```
Stillness-2: Using `[baseImageUrl stringByAppendingString:[metadata valueForKey:@"image"]]` without checking result of the `[metadata valueForKey:@"image"]` is dangerous.

Geekfish: I'm sorry, I'm not usually an Objective-C developer, can you explain what you mean by "dangerous"?
I copied this sort of logic from the existing media strategies, for example: https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/YouTubeStrategy.m#L68

Geekfish: Unless this is specific to string concat, but I would appreciate clarification regardless :)

Coder-256: @Stillness-2 I fixed this problem in #340.

Geekfish: @Coder-256 I don't understand how this fixes the underlying issue of the insufficient selector. I've yet to encounter a track without an image element, so the solution is probably not about hiding generic exceptions under the carpet?

Geekfish:
I have updated this PR with a fix for the issue @Stillness-2 encountered.
@Stillness-2 I would appreciate some clarification on the what you consider dangerous code so I can also fix this asap.

BeardedSpice/MediaStrategies/RadioSwissJazzStrategy.m
Geekfish: @Coder-256 I have added a commit that fixes said selector issue.

Coder-256: @Geekfish I may have misunderstood. I thought the issue was that there wasn't always album art, which would cause the code to throw an exception, so I made it fall back to an empty string. What is the problem, if not that?

Geekfish: Also waiting for an explanation on this, but in terms of having a missing element, I think the image element is kind of "guaranteed", I would expect, as much as the title or artist name (I've been listening to this radio for a long time :older_woman:). Also the existence of the element != the existence of a valid src... maybe it's related to that.

Coder-256: @Geekfish It seems that the problem was resolved in ac1cc6fc5b88fd4f46d65738f8f32f0445a3272d.

Stillness-2:
Only this:

`(NSString *)stringByAppendingString:(NSString *)aString`
aString - The string to append to the receiver. **This value must not be nil.**

**Raises an NSInvalidArgumentException if aString is nil.**

But now that doesn't matter because I put try/catch on the method call

Stillness-2:
Also you must implement `pause` method

Stillness-2:
@Geekfish I asked you to implement `pause` method some time ago.
https://github.com/beardedspice/beardedspice#about-pull-requests

Geekfish:
@Stillness-2 sorry, forgot to push. Thanks for the feedback.

PR #333, opened 2016-02-07
Add iHeart Radio Strategy

Coder-256:
Resolves #326 and #258.

Note: `previous` method is left out intentionally, as iHeart Radio does not support it.

Issue #258, opened 2015-10-16
[App Support] http://www.iheart.com I Heart Radio


bastiandantilus:
Just realized there is no http://www.iheart.com I Heart Radio support.

Is that a possibility?

1ps0:
Added in df1fd25c95da6910d553634d2392e1ef1b2e3e9e

PR #337, opened 2016-02-13
Add iHeart Radio Strategy – try #2

### LINKED ISSUE #326: [App Support] BeardedSpice
Can you please add iHeartRaido to the website and maybe app controller? Thanks!

Coder-256:
Same as #333, but resolves branch issues. Use this instead of #333. This also resolves #326 and #258.


Stillness-2:
does not work in sections: Artist Radio, Podcasts.. and may be some others :(

Stillness-2:
here's, how to trust people :)

Coder-256:
Sorry, I only tried it in Live Radio. It didn't occur to me that it wouldn't work in other sections of the website. I'll test better next time. I fixed the problem in e34bb782.

PR #339, opened 2016-02-15
GoogleMusic and Songza updates

### LINKED ISSUE #317: Improper whitelisting for Google Play
It seems that the implementation for Google Play Music isn't using a specific enough URL, as it is detecting this tab for Google Play Books as a controllable option (see screenshot). I've tested using the media controls and they do not affect the page adversely. It's just unexpected and probably unnecessary if it doesn't actually do anything. 

screen shot 2016-01-13 at 4 38 32 pm

It also seems to incorrectly detect Google Play Store, Apps, Newstand, Devices, but does not work on Google Play Movie.

fmborghino:
- Songza site is dead, absorbed into Google Music as of 2016-02-01,
  removed strategy
- Fixed Google Music title + album art
- Added Google Music isPlaying, favorite
- Narrowed Google Music url selector, fixes #317

I miss Songza.


Stillness-2:
You have write access. Why do you make PR? :)

Coder-256:
+1 Songza was great :(

fmborghino:
> Why do you make PR?

Heh, habit. Also I figured it wouldn't hurt to let it sit out there for a day in case there were any comments.

I'll get it merged :)

PR #352, opened 2016-03-05
Update Wonder FM strategy to work with Wonder 2.0

kconarro14:
Wonder.fm launched a redesigned site on 2/17/16. They also launched the Primary (electronic) and White Label (hip hop) sites. This PR updates the Wonder strategy to work with the new site design.

See #359 for the additional support for Primary and White Label strategies.

### DIFF
--- a/BeardedSpice/MediaStrategies/WonderFmStrategy.m
+++ b/BeardedSpice/MediaStrategies/WonderFmStrategy.m
@@ -26,13 +26,13 @@
 }
 
 - (BOOL)isPlaying:(TabAdapter *)tab{
-    NSNumber *result = [tab executeJavascript:@"(function(){return $('div.jp-audio').hasClass('jp-state-playing');})();"];
+    NSNumber *result = [tab executeJavascript:@"(function(){return document.querySelector('.track--active:not(.track--paused)')})();"];
     return [result boolValue];
 }
 
 -(NSString *) toggle
 {
-    return @"!function(){var e=document.querySelector('.jp-type-single'),l=document.querySelector('a.jp-play'),t=document.querySelector('a.jp-pause'),c=document.querySelector('.track_play'),u='none'===getComputedStyle(e,null).display,n='none'===getComputedStyle(l,null).display;u?c.click():n?t.click():l.click()}();";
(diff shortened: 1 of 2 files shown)


fmborghino:
This is a nice update - the js seems to work fine. There's a couple of items to handle:

- Add these 3 to the README.md, sort them into the list there
- You should also override the displayName method in `WhiteLabelStrategy` and `PrimaryFmStrategy' else the preferences show `Wonder FM` for all of them. Also this is a display name, so it's ok to give it spaces if that looks more natural (your choice). For example:

```
-(NSString *) displayName
{
    return @"Primary FM";
}
```

Finally I notice a bug that I don't yet understand - if I open all 3 sites, they don't behave well in the app icon's menu. They come and go. Not sure what that's all about. Can anyone else repro that bad effect?

Coder-256:
@fmborghino 

> if I open all 3 sites, they don't behave well in the app icon's menu. They come and go.

That's probably due to the issue you mentioned about the `displayName`.

fmborghino:
I don't think so... I fixed the names in a local branch. Same problem.

kconarro14:
@fmborghino made the readme + `displayName` updates, thanks for pointing those out. Rebased as well.

As for the issue with running multiple sites, I _am_ able to reproduce. I'll try to do some more digging, let me know if you figure anything out in the meantime.

fmborghino:
I can't figure out what's going on with these sites. Any two of them causes this problem. I thought perhaps there's a problem with how we're handling the strategy classes if the inheritance is in place, but this still happens even if you go with straight up duplication of everything.

Here's a suggestion tho'. How about you change this PR back to the Wonder FM update - that works fine and gets that site working. Then do a new PR with Primary FM and White Label FM and we can solve that sub-problem.

Or if you're ok with the hold-up on solving the multi-site problem, leave this be. I'll have another look when I get time, and someone else may have the key insight. This has to be something simple.

kconarro14:
@fmborghino I'll separate the PRs to get Wonder working again. Thanks for troubleshooting, hopefully we can get to the bottom of it.

kconarro14:
@fmborghino updated this to be Wonder only, added #359 to handle the 2 new sites.

fmborghino:
Works great for me, thanks for splitting this up.

Issue #347, opened 2016-02-27
[App Support] VLC Media Player


zakiaziz:
VLC is a pretty ubiquitous  tool for audio. I personally use it to stream podcasts and DI.fm. It'd be cool to see VLC support

PR #356, opened 2016-03-07
Add VLC app support

fmborghino:
Putting this up for comments as I've never done an app adapter.

The scripting model for VLC has multiple windows and documents, which makes sense. My approach to finding the right object to receive the controls seems hack-ey - I compare the document names against the app's idea of the currently selected media name. Seems to work, but it doesn't make me happy.

@Coder-256 I think you were playing with this too (noticed your question on SO) and I think I ran into exactly the same problems. See what you think of this approach.

@Stillness-2 I think I may have missed something about the API, is there a better way I should be looking through an app's multiple windows? (key() and isEqual() worry me, I should probably be doing something with those eh?) :)

Might address #347

### DIFF
--- a/BeardedSpice/Tabs/VLCTabAdapter.m
+++ b/BeardedSpice/Tabs/VLCTabAdapter.m
@@ -1,0 +1,141 @@
+//
+//  VLCTabAdapter.m
+//  BeardedSpice
+//
+//  Created by Max Borghino on 2106-03-06
+//  Copyright (c) 2015 Tyler Rhodes / Jose Falcon. All rights reserved.
+//
+
+#import "VLCTabAdapter.h"
+#import "VLC.h"
+#import "runningSBApplication.h"
+#import "NSString+Utils.h"
+#import "MediaStrategy.h"
+
+#define APPNAME         @"VLC"
+#define APPID           @"org.videolan.vlc"
+
+@implementation VLCTabAdapter
+
++ (NSString *)displayName{
+
+    return APPNAME;
+}
+
++ (NSString *)bundleId{
+
+    return APPID;
+}
+
+- (NSString *)title {
+
+    @autoreleasepool {
+        VLCApplication *vlc = (VLCApplication *)[self.application sbApplication];
+        NSString *title;
+        if (![NSString isNullOrEmpty:vlc.nameOfCurrentItem])
+            title = vlc.nameOfCurrentItem;
+
+        if ([NSString isNullOrEmpty:title]) {
+            title = NSLocalizedString(@"No Track", @"SpotifyTabAdapter");
+        }
+
+        return [NSString stringWithFormat:@"%@ (%@)", title, APPNAME];
+    }
+}
+- (NSString *)URL{
+
+    return @"VLC";
+}
+
+// We have only one window.
+- (NSString *)key{
+
+    return @"A:VLC";
+}
+
+// We have only one window.
+-(BOOL) isEqual:(__autoreleasing id)otherTab{
+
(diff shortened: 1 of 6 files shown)


fmborghino:
Oh also @Coder-256 thanks for writing up how to add an app in the wiki. Very handy.

BeardedSpice/Tabs/VLCTabAdapter.m
```diff
@@ -0,0 +1,141 @@
...
+// NOTE: this feels brittle, it seems there should be a better way
+- (VLCDocument*) activeDocument {
+    VLCDocument *doc = NULL;
+    VLCApplication *vlc = (VLCApplication *)[self.application sbApplication];
+    if (vlc) {
+        NSEnumerator *e = [[vlc windows] objectEnumerator];
+        id o;
+        while ((o = [e nextObject]) && doc == NULL) {
+            NSLog(@"VLC window %@: %lu: %@", o, [o index], [o name]);
+            if ([[o name] isEqualToString: [vlc nameOfCurrentItem]]) {
```
Stillness-2: pathOfCurrentItem ?

Stillness-2:
> (key() and isEqual() worry me, I should probably be doing something with those eh?)

Nativapp interface is designed to work with one window at the moment :) So all is correct. :)

BeardedSpice/Tabs/VLCTabAdapter.m
```diff
@@ -0,0 +1,141 @@
...
+        [doc play];
+    }
+}
+
+- (void)pause{
+    VLCApplication *vlc = (VLCApplication *)[self.application sbApplication];
+    if (vlc && [vlc playing]) {
+        VLCDocument *doc = [self activeDocument];
+        if (doc) {
+            [doc play];
```
Coder-256: @fmborghino You guessed right. [In my question](http://stackoverflow.com/q/35670703), I was trying to use `[vlc stop]` to pause, which was bad for a number of reasons:

1. It actually closed the file.
2. This caused it to hang when resumed.
3. For some reason (bug?), `[vlc stop]` is completely unavailable, along with some other methods. However, it is available [by using plain Applescript](http://stackoverflow.com/a/35774824/3398839), which I was [trying to avoid](http://stackoverflow.com/questions/35670703/applescript-vlc-scripting-bridge#comment59241547_35774824).

It never occurred to toggle it with `play` (most applications like VLC have separate play and pause methods). Also, my implementation only worked if there was one window. I admit, you beat me to it with a much better implementation. :disappointed:

fmborghino: Ah yeah those are the problems I ran into, my understanding so far (and bear in mind I don't touch Obj-C much at all) is that those methods are an `@protocol` on other objects than the application itself. Specifically they end up mixed-in on the windows and the documents. Hence the whacky code to search the docs for the right one! This was all a lot of trial and error, so whilst it works, I am not super happy about it. Oh well.

BeardedSpice/Tabs/VLCTabAdapter.m
fmborghino: Yeah I thought about that, and it seems like it might be better (less likely to conflict with anything else), but the window doesn't have a path, and I can't seem to get the document to show me the path either. I suspect this is pilot error.

The  current scheme (app's name of current item == window's name) seems to work ok.

fmborghino:
Tested this some more, seems to work with anything I can put into VLC's playlist with the various "open" operations. Local music, video, mounted DVDs, URLs.

PR #358, opened 2016-03-07
Fixed spelling error (demon -> daemon).

Coder-256:
The variable `remoteControlDemonEnabled` should be spelled `remoteControlD`__`a`__`emonEnabled`.


fmborghino:
Heh ok. I was going to argue that the usage is pretty split but I changed my mind after checking all the `man` pages on my local box:

```
$ man -k '.'|grep -i demon|wc -l
       0
$ man -k '.'|grep -i daemon|wc -l
      87
```

Fair enough :+1:

PR #359, opened 2016-03-08
Add Primary FM and White Label FM

conarro:
Originally part of #352, but split it out as we ran into issues when 2 or more sites were open at once (any of Wonder, Primary, White Label).

Once #352 is merged I can rebase this and we can continue debugging, with Wonder support getting fixed in the meantime.


kconarro14:
rebased

1ps0:
Uh, where's the meat? 

You have a basic recognition regex, but no javascript to apply to any of the functions, including the most important one, `toggle`.

1ps0:
Yea, there's no meat here. You have basic regex to identify the domain, but no javascript to perform actions. Without the javascript this isn't a complete strategy. Please review and update, preferably with a pure javascript strategy (see master README).

conarro:
@1ps0 These 2 sites are built on the same platform as Wonder.fm, so the thinking was that the logic could be shared. The strategy was implemented in #352, with these strategies inheriting from [the Wonder strategy](https://github.com/conarro/beardedspice/blob/master/BeardedSpice/MediaStrategies/WonderFmStrategy.m) to avoid duplicating logic.

If there's a preferred way to do this (different from that already in this PR), let me know and I'd be happy to look at tweaking it.

As for the general usage issues/conflicts we've seen when running these strategies, I haven't gotten around to debugging further.

1ps0:
Ah ok. That makes sense. 
We're looking at other ways to extend javascript strategies, this would definitely benefit from that.
For now though, simply copying the javascript should be fine.

Stillness-2:
Hmm. If both sites have the same js interface, you may, simply,  make appropriate predicate.

1ps0:
This code has been converted and imported into the project as of earlier this year. 

Closing this PR accordingly.

Issue #363, opened 2016-03-20
[App Support] Dailymotion


Kureb:
It would be nice to have support for http://www.dailymotion.com/

PR #364, opened 2016-03-20
[App Support] Dailymotion

Kureb:
Play / Pause / Show Info / (un)like are working for playlists and single videos.
No controller for next nor previous when playing a playlist because no button are available on the original website.

Note that you have to use HTML5 version of dailymotion (read the comments in DailymotionStrategy.m)!

References to #363


fmborghino:
@Kureb thanks for implementing. Seems to work, except for your `paused` method. The interface requires a `pause` method, and this figures out how to pause the player. So typically, do nothing if not already playing, or hit pause (or stop if that's the only thing available) if playing. Looks like you have all the logic there to handle this. Please implement.

This is used, for example, (good way to test) in the scenario where your player is playing, and you have another tab with a different supported player open, but not playing. You then select that other player from the BS menu. This should cause your player to pause.

Once you fix that, I think we're good to go.

Kureb:
@fmborghino Hi, thanks for feedback! It should now be fixed :)

fmborghino:
OK! Everything seems to work for me now. Thanks.

PR #373, opened 2016-04-04
Added TurkTelekom Muzik

Binusz:
I did it for my friend. It should be next release :)


1ps0:
@Binusz since this is region bound for Turkey, you're going to have to keep an eye on this strategy.

Binusz:
What you mean ? i don't understand. if you mean that the TurkTelekomMuzik just serve in Turkey region, This domain can accessible anywhere...

1ps0:
Yes, I can see the site player and test the javascript, but that is all, no music. If it needs to be fixed in the future, we will need your help :)

Binusz:
I see. :) 
that web site has very strange playing method. Firstly, you have to make a playlist then you need to play music...

Coder-256:
I get a (partially covered) message saying:

> Dear Türk Telekom Müzik Visitor,
> Your IP is &lt;redacted&gt;
> We are sorry to say that due to licensing constraints, we can not allow access to visitors located outside of Turkey.
> If you believe we have made a mistake about your location, we apologize and ask that you please dial 444 0 375 for customer services.
> We share your disappointment and greatly appreciate your understanding.
> Sincerely,

The message ends there.

I can play 30 second previews of songs. I tried creating an account and logging in (which was difficult considering I know no Turkish and that Google Translate was down; I eventually figured it out, though). It still gave me that message and would only let me play previews, though.

PR #377, opened 2016-04-06
Update Subsonic strategy for new player in Subsonic 6.0+

tstackhouse:
In the newest (beta) version of Subsonic 6.0, the JWPlayer was replaced with what appears to be an HTML5-based player.  The main change was we were directly interfacing with the JWPlayer API for play/pause, and using the JS API for other actions.  This should be backwards compatible with the older versions that still use JW while also working with the new player.

### DIFF
--- a/BeardedSpice/MediaStrategies/SubsonicStrategy.m
+++ b/BeardedSpice/MediaStrategies/SubsonicStrategy.m
@@ -26,13 +26,22 @@
 
 - (BOOL)isPlaying:(TabAdapter *)tab
 {
-    NSString *playerState = [tab executeJavascript:@"window.frames['playQueue'].jwplayer().getState()"];
+    NSString *playerState;
+    NSString *isNewPlayer = [tab executeJavascript:@"typeof window.frames['playQueue'].localPlayer"];
+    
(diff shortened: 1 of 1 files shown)


tstackhouse:
XCode is a butt with editing plist files... Let me know if I should change the formatting.

askedrelic:
Nice backwards compatible JS! I just upgraded to Subsonic 6.0 and was testing out the new API, so I will verify this works for me.

1ps0:
Sorry to pull the rug out from under you. We did one last (final-ish) change to the strategy format. When you have a moment can you swap it out? This strategy will be merged in as part of #429.

alexfornuto:
Any love for this PR? I'm hoping this will fix BS compatibility with Libresonic 6.1

1ps0:
I think I tried to convert it to the JS strategy format a little while ago and ran into some roadblocks. I'll take another look this weekend.

radiosilence:
Did anyone ever do this?

tstackhouse:
I'm not on a Mac any more, sadly, so I'm of no more help, but it looked like @1ps0 took up the mantle.

PR #379, opened 2016-04-09
Fix Plex tabs

askedrelic:
Plex just updated with a new React based webapp: https://blog.plex.tv/2016/04/07/seek-plex-shall-find-leveling-web-app/

This fix works for me locally and I think should be backwards compatible, but I can't really verify since I've upgraded my Plex server.

### DIFF
--- a/BeardedSpice/MediaStrategies/PlexWebStrategy.m
+++ b/BeardedSpice/MediaStrategies/PlexWebStrategy.m
@@ -13,7 +13,7 @@
 
 - (BOOL)accepts:(TabAdapter *)tab {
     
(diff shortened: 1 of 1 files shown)


1ps0:
This probably should be created as a second strategy for plex specific to the major revision. I say this because as a locally hosted website other people will still be running and using older versions.

1ps0:
Will merge this as part of #429. Can you take a moment and convert it to the new javascript strategy format? :)

askedrelic:
FYI I've looked at this and will be trying to updated this PR for #429 but anyone can update it, if they get there first.

askedrelic:
Closing this, should be fixed here: https://github.com/beardedspice/beardedspice/pull/379

PR #382, opened 2016-04-13
Versioned strategy

1ps0:
Details extended from #302 

Addressed comments by @Stillness-2 in the original PR. Ready for review #2


Stillness-2:
Why do you create PR? You may push to repository. We have branch for this task. :)

Stillness-2:
What do you think about: _"Having js as strings in Obj-C is no fun. Having it in plists is no better. If we are going to go this route, we could also consider a direction (sure, future) where the strategy can be handled as a plain .js file with all the methods, or a directory with several js files per method. We can talk about the details. Something that is more native for strategy coders in js would make it easier in general."_?

May be js file with:
```
strategy = {
'isPlaying' = function(){..},
'pause' = function(){...},
...
}
```
or something like that?

1ps0:
We could do something like that. However, the pain points end up at the parser. 

I think I picked plists over JSON or similar formats because it was so easily integrated into the existing Obj-c code. The stack is already built in (network and parsing primarily). Anything else would require more moving parts (read: third party libraries), and would make integration more complicated and require more testing. Though, this would be the best time to re-examine it, and the changes required--in theory--would not be very large. 

I also looked into splitting up into multiple files per strategy but it got cumbersome and bloated very quickly. For the short and medium foreseeable future a single file per strategy is best.

And since I was working with you to integrate this code, I wanted to give you a chance to peer review it before I pushed.

1ps0:
Also, I'm assuming that I would only be merging into master, and that you are the only developer currently doing releases?

Stillness-2:
@1ps0 "However, the pain points end up at the parser."

Stillness-2:
@1ps0 "However, the pain points end up at the parser."
Check this:
https://github.com/AdguardTeam/AdguardForiOS/blob/master/AdguardExtension/SharedComponents/Services/AESFilterConverter.m

`init` method:
```
#ifdef DEBUG
        else{
            DDLogInfo(@"ConvertFunction: \n%@", [_converterFunc toString]);
        }
#endif
```
`[__converterFunc toString]` gives us string representation.

Where https://github.com/AdguardTeam/AdguardForiOS/blob/master/AdguardExtension/SharedComponents/Services/JSConverter.js is a pure JS file.
However, we must increase deployment target to 10.9 :)

1ps0:
Ah ok. I see where you're going with that. However, it counts as a third party library and any bugs in that code would become bugs in beardedspice. If you want, we can work to bring a copy into this project, but I would suggest it be a separate PR. Right now I'm following the KISS principle for such a large refactor of this project.

I'm going over the code one last time. Is there anything you want to add before I merge it? Or do you want to have a serious discussion about the JSConverter code?

Stillness-2:
"Or do you want to have a serious discussion about the JSConverter code?" NO. I simple shown you approach, which gives ability to use JS file for strategy, reading, compiling, getting functions code from it. There is used standard JavaScriptCore framework: https://developer.apple.com/library/tvos/documentation/Carbon/Reference/WebKit_JavaScriptCore_Ref/index.html
http://nshipster.com/javascriptcore/

BeardedSpice/AppDelegate.m
```diff
@@ -388,23 +382,82 @@ - (void)headphoneUnplug{
...
+
+static NSString *const kBSUpdateCheckingTitle = @"Checking...";
+
+- (IBAction)checkForUpdates:(id)sender
+{
+    // quietly exit because this shouldn't have happened...
+    if (statusMenu.numberOfItems == 0)
+        return;
+
+    NSMenuItem *item = [statusMenu itemAtIndex:0];
```
Stillness-2: Why first item of the menu?

Stillness-2:
Also, we need disable menu item "Check for Compatibility Updates" while upgrade is performed

1ps0:
Ok, rebased a fixup that disables the check updates when selected, and finds the menu item by tag number set in `MainMenu.xib`. Not sure why I didn't finish writing that. 

Tested accordingly. Had to use the Network Link Conditioner to properly see ;)

Stillness-2:
And one moment.. Please, you must use `NSLocalizedString` for defining string constants. May be we will be support localizations in future. :)

1ps0:
haha sure why not. I had the same thought, but I figured I'd do it as we moved in that direction. Localization can get messy easy ;)

1ps0:
Hokay. Basics setup.

Coder-256:
@1ps0 @Stillness-2 Don't rule out JSON. See the [`NSJSONSerialization` class](https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSJSONSerialization_Class/).

1ps0:
@Coder-256 
I gave it a quick prototype and ran into the problems I was expecting. In short, the JSON spec is too rigid for our needs. We would have to do it in js which returns us to our original discussion about JSCore.

@Stillness-2 
Thanks for clarifying how you want localized strings to be done. I'll follow that pattern in the future. Anything else you wanna throw in before I pull into master?

1ps0:
One final test and all looks green. Going to be pulling this in to master shortly. 

@Stillness-2 I'm assuming you're doing the actual binary deploy?

Stillness-2:
"I'm assuming you're doing the actual binary deploy?" No. I suggest to make a new release only after #383 . This will be a more complete solution& :)

PR #390, opened 2016-04-24
Improved template explanation

Coder-256:
Also provided an example for `trackInfo`.

### DIFF
--- a/template-explained.plist
+++ b/template-explained.plist
@@ -44,7 +44,7 @@
     - Normal formatting is supported (can copy/paste with newlines and indentations)
     - &amp; is used to escape '&' so the file is readable.
     -->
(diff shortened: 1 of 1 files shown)


1ps0:
As much as I love the increased detail, I dont think this is going to be pulled in. It seems we're waiting on my JSCore conversion before releasing the strategies revamp, so this will get blown away when I change it out. 
@Stillness-2 can you confirm?

Stillness-2:
yes

1ps0:
@Coder-256  I've added these changes to the upcoming branch js-conversions. If you could look there to see if you'd like to make any other additions like this? 

Also can you close out this PR. :)

PR #406, opened 2016-05-20
Added KCRW strategy 

geexican:
Strategy for integrating BeardedSpice with KCRW.com

toggle, previous, next, and pause all implemented, although pause is the same as toggle


1ps0:
Merging this in shortly. Also doing the conversion from plist to js.

Issue #385, opened 2016-04-16
[App Support] ProductHunt Podcast Player


yuyangchee98:
Hey, would love it if you guys could support ProductHunt's podcast player. Love your product by the way. Thanks for everything.

PR #409, opened 2016-05-23
Add ProductHunt strategy

Kureb:
As asked in issue #385

- toggle
- previous
- next
- favorite
- pause
- isPlaying
- trackInfo


Kureb:
I edited my comment because the referenced issue is wrong, #385 instead of 395 (my bad, sorry.. I should have double checked)
But somehow the change is not persistent

Coder-256:
It seems that you also regenerated the project to support a newer version of CocoaPods (which probably happened automatically).

1ps0:
Merging this in shortly. Also did the conversion from plist to js

PR #420, opened 2016-06-04
Js conversion cleanup

Stillness-2:
@1ps0  This is what I meant about strategies update. Maybe there bugs. :)


1ps0:
Out of curiosity, what tool do you use for git?

Stillness-2:
https://desktop.github.com

PR #423, opened 2016-06-07
Js conversion

1ps0:


1ps0:
@Stillness-2 Aside from adding any dev-friendly functions, I think this covers everything that needs to be added.

For this PR we have:
- Conversion of all strategies to Javascript
- Removal of local versions.plist in favor of scanning (1) bundle, (2) saved folder in Application Support, (3) custom folder in Application Support. Where (3) has the highest priority when compiled as DEBUG.
- Added validation tests at git push time to make sure everything that goes into the repo passes basic syntax and similar checks.

Miss anything?

Stillness-2:
@1ps0 What about deployment target? Do we raise to 10.9?

1ps0:
Done, and updated against master. Should be conflict free.

Stillness-2:
I still need a time to finish the review and tests :)

1ps0:
No rush. Just wanted to wrap it up enough so you could do exactly that :)

Stillness-2:
@1ps0 Please check last changes.

Stillness-2:
I think in this state we may merge to master

1ps0:
Hmm. Normally I'd prefer the BSMediaStrategy to be built and retained and updated in one object, because it retains the data of the file location, and can be reloaded quickly. Especially in development. 
Was there a particular reason for going this way? 

Especially with the copying. Why not just use the way apple provided us. :)

```
    // this is basically handing off the retain to this object, cause strategy is gonna go get garbage collected.
    self.strategyVersion = [strategy strategyVersion];
    self.strategyURL = [strategy strategyURL];
    self.strategyJsBody = [strategy strategyJsBody];
    self.custom = [strategy isCustom];
    self.fileName = [strategy fileName];
    self.scripts = [strategy scripts];
    self.acceptParams = [strategy acceptParams];
```

The only time I wouldn't use `self.propertyName` to assign stuff is in the initializer because of the same side effects we want in the above snippet. [See discussion](http://stackoverflow.com/a/8056260).

1ps0:
To be more precise...

I was aiming for BSMediaStrategy to be a concrete per-strategy singleton, with the repository being in BSStrategyCache. This means that an object would be created for every strategy found, even if its invalid. The isLoaded property is critical for this because it allows each strategy to do internal state checking and updating. Otherwise it might as well be a struct, right?

Stillness-2:
The object remains the same. While copying of state, the object reference does not change. 

"Was there a particular reason for going this way?" 
This is a simple way to keep the old object if any error occurs while loading from new path.

Stillness-2:
Using self.strategyVersion = [strategy strategyVersion];
    or _strategyVersion = strategy-> _strategyVersion;
In this case, no matter what you choose . just a second variant is a little bit faster .

Stillness-2:
@1ps0 "This means that an object would be created for every strategy found, even if its invalid."
What are the advantages of this approach?  Why do we must hold in the cache non-working strategy?

1ps0:
For the copy function, you're doing pointer operations (which can be hard to understand) instead of simpler and cleaner operations to save nanoseconds. We need some more optimization, but that's not one of the places.

For `BSMediaStrategy`, my first reaction was related to design. I had a "hey, I'm using that" moment. 
I have a longer term plan to increase feedback to the user. Looking at the types of Issues that we are getting, this seems more and more necessary to me. 
A simple example might be to show in the GeneralPreferences panel which strategies failed to load (instead of currently just removing them from the list). Similarly, by keeping a reference and showing when a strategy is fixed, we can let the user know on update. 

For actual design though, building an new object only to copy out it's guts, goes against object oriented programming. You're using its private internals to do an external operation. 

Normally I'd expect the design to look like one of these:
- You could have a `BSMediaStrategyUpdater` class and do Dependency Injection to update it (this is the Java way. Overkill in this case).
- You could create a new `BSMediaStrategy` object and replace all the existing references from the old object to this new object (but it gets messy to make sure all references are replaced).
- You could allow `BSMediaStrategy` to manage itself, since the logic is simple and the data and operations are extremely constrained and standardized.


The direction you're going in makes sense, but I want to make sure that my future self, as well as new devs, can easily build on these components :)

Coder-256:
Why 17ef6e6? Even if the method is private, you shouldn't remove its documentation. Documenting code makes it easier for everybody to use.

Stillness-2:
I did not remove documentation :)

Coder-256:
Oh, I didn't realize that you just moved it.

1ps0:
@Stillness-2 your commit messages are hilarious.

Anyway, not exactly what I was talking about, but at this point i'd say good enough. If its a problem later I'll change it later.

*Official stamp of approval*

PR #450, opened 2016-07-16
Fix plex toggle strategy

askedrelic:
I jumped the gun a bit on https://github.com/beardedspice/beardedspice/pull/449

This PR fixes the toggle functionality, which has been broken I think in master?

I'm on Plex Web 2.7.0 (latest) and I tested all major functionality
- Music/video play/pause work
- next/previous with music work
- all music/video info appears correctly in the menubar info

### DIFF
--- a/BeardedSpice/MediaStrategies/PlexWeb.js
+++ b/BeardedSpice/MediaStrategies/PlexWeb.js
@@ -26,7 +26,7 @@ BSStrategy = {
     var thePlayer = document.querySelector('.player.music') ? '.player.music' : '.video-player';
     var pauseButton = document.querySelector(thePlayer + ' .pause-btn');
     var buttonIsHidden = pauseButton.classList.contains('hidden');
-    var toggleButton = document.querySelector((buttonIsHidden ? ' .play' : ' .pause') + '-btn');
+    var toggleButton = document.querySelector('button' + (buttonIsHidden ? '.play' : '.pause') + '-btn');
     toggleButton.click();
   },
   next: function () {
--- a/BeardedSpice/MediaStrategies/PlexWeb.js
+++ b/BeardedSpice/MediaStrategies/PlexWeb.js
@@ -6,7 +6,7 @@
 //  Copyright (c) 2015 BeardedSpice. All rights reserved.
 //
 BSStrategy = {
-  version:2,
+  version:3,
   displayName:"Plex Web",
   accepts: {
     method: "script",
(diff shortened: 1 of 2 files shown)


bcomnes:
Pausing works again, resuming (hitting play again) seems to restart the playlist.

bcomnes:
Fixed the resume functionality, see here: https://github.com/askedrelic/beardedspice/pull/1  However you guys want to integrate that is fine with me.  

Does the version # need to be bumped to 3?

1ps0:
Yes we need to raise the version. Every time the strategy file is updated the version should be updated in the strategy file and its versions.plist entry.

Because this is an installed versioned app, not a universally updated website, we should move this to be version specific. 

Instead of replacing the old version can you put this in a new file specific to the version? Say for example `PlexWeb-2-7.js`. Do we know for sure it was 2.7 that changed the controller javascript? 

Similarly the accept script should be changed in both versions of the PlexWeb strategy to tell BS which one to use (say for example the version number is on the page somewhere). The approach that comes to mind says start with the 2.7 strategy and if the discovered page version is below that use the one before.

bcomnes:
I don't know for sure if the change was introduced in 2.7, and given the change I had to introduce on top of this PR, it seems like they are willing to make dom changes on the reg.

While libraries can serve the web app at whatever version it has installed, clients only need to visit https://plex.tv/web/ to stream, where the latest web app is served up, and I don't think Plex supports any version of their web app other than the latest.  I would advocate that the Plex strategy only aims to support the latest Plex web, and let individuals float their own custom strategies if they refuse to update their plex libraries.  Thoughts? 

It seems like @askedrelic might not be available right now.  Should I point my PR (which includes this PR + my changes) directly at the BS, and bump the version number?

1ps0:
If he's out for the moment, go ahead and I'll pull it in. 

We can deal with the legacy versions of installed apps at a later time. It may end up having compiled code changes.

bcomnes:
Woo! Thanks @askedrelic 👍

askedrelic:
I don't have a strong opinion on either way, but I think supporting only latest Plex shouldn't cause many problems:
- the platform doesn't move that quickly 
- the changes for 2.7 seem like the first time Plex has "broken" their "API" in years. Most users seem to be running latest Plex (I do atleast)

1ps0:
It's good enough for now. Something we'll address in a future update when the need becomes more glaring.

Thanks! All merged :)

askedrelic:
So cool that is updates automatically and I already have it locally!

1ps0:
I know right!? Totally worth all the hard work :)

bcomnes:
Yes bravo!

PR #464, opened 2016-08-05
brighter menu bar icons

bolandross:
increased brightness of menu bar icon, so it matches the colors of OSX El Capitan and custom apps like Dropbox, etc.

I rendered three resolutions (default, 2x and 3x)out of the original pdf file and saved them as PNGs. I kept the original vector PDF so future changes or corrections can be made. I am not a photoshop pro, so maybe there will be somebody in the future who can do a better job than me.


Stillness-2:
I don't like gradient in this place. I suggest to use #181818 as base colour. Why do you use png format?

bolandross:
Because I just wanted to help and didn't know better.

PR #465, opened 2016-08-07
Support white plastic Apple Remote

ytsutano:
The original white plastic Apple Remote has a different cookie string for the play/pause button.
- Play/Pause Button:
  - White Plastic: `33_23_21_20_2_33_23_21_20_2_` (Added)
  - Aluminum: `33_21_20_8_2_33_21_20_8_2_`
- Sleep Button:
  - White Plastic: `37_33_21_20_2_37_33_21_20_2_` (Added)
  - Aluminum: `35_31_19_18_35_31_19_18_`

Tested on OS X El Capitan 10.11.6.

### DIFF
--- a/BeardedSpiceControllers/Controllers/DDHidLib/DDHidAppleRemote.m
+++ b/BeardedSpiceControllers/Controllers/DDHidLib/DDHidAppleRemote.m
@@ -113,6 +113,8 @@
                                   forKey: @"18_14_6_5_18_14_6_5_"];
     [mCookieToButtonMapping ddhid_setInt: kDDHidRemoteButtonPlay_Sleep
                                   forKey: @"35_31_19_18_35_31_19_18_"];
+    [mCookieToButtonMapping ddhid_setInt: kDDHidRemoteButtonPlay_Sleep
+                                  forKey: @"37_33_21_20_2_37_33_21_20_2_"];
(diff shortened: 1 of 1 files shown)


1ps0:
@ytsutano  What HID remote did you test with? Was it [this one](http://www.apple.com/shop/product/MM4T2AM/A/apple-remote)? And does it require an IR port?

ytsutano:
@1ps0 No, that's the aluminum remote from 2009 which is already supported. My patch is for the [whilte plastic one](https://en.wikipedia.org/wiki/File:Apple_Remote.jpg) from 2005.

And yes, it does require an IR port.

1ps0:
Do you have a way to test without an actual IR enabled mac or remote?

ytsutano:
@1ps0 I've tested it on iMac (10.11.6 El Capitan) and MacBook Pro (10.12.1 Sierra) with an IR receiver. On my newer Macs w/o the receiver, it does nothing.

I don't have a way to test this on the newer Macs or without the remotes, unfortunately. This patch is simple as it only adds two entries without any control flow changes, so it shouldn't cause any issues. I guess the only way is to take my word for it, or let others test.

1ps0:
@ytsutano Ok, this was a bit of a beast to test. Gonna merge it in, and I think that'll be the whole of the `2.1.1` release. 

Thanks for bearing with me :)

Issue #355, opened 2016-03-06
[App Support] Downcast


filnko:
That would be really nice! Sadly there's no information on how to create a MediaStategy for native apps.

Coder-256:
[Now there is!](https://github.com/beardedspice/beardedspice/wiki/Native-App-Developers'-Guide)

fmborghino:
@Coder-256 this is really nice! Thanks.

I don't know much about scripting support, but it seems not all apps include this. One example player is SomaFM.app.

Is it worth adding a 3.1 item saying that if you see something like the following, you are out of luck? ...

```
$ sdef /Applications/SomaFM.app/ | sdp -fh --basename SomaFM
sdef: couldn't get sdef for /Applications/SomaFM.app/ (error -192)
sdp: stdin - zero length data
```

Not sure this is worth adding to the wiki, perhaps it's too obvious. Just an observation.

Thanks again. I think I'll see about adding VLC as an exercise!

Coder-256:
Yeah, you could add that. If that is the output, then that application is not scriptable. See this: <https://developer.apple.com/library/mac/documentation/AppleScript/Conceptual/AppleScriptX/Concepts/scriptable_apps.html>

fmborghino:
Done. Thanks for the pointers.

Coder-256:
@filnko Could you run the following commands and post the output of the commands and the resulting file (Downcast.h) from your Downloads folder (if it exists)?:

```
cd ~/Downloads
sdef /Applications/Downcast.app | sdp -fh --basename Downcast
```

I would do it myself, but I do not have Downcast and do not want to buy it.

filnko:
sdef: couldn't get sdef for /Applications/Downcast.app (error -192)

Travmatth:
I emailed Downcast today and mentioned this issue, if they reply I will update.

gcox:
I'll be looking into adding support for this in [Downcast](http://downcast.fm) soon. Thanks for making me aware of it, @Travmatth.

1ps0:
This will be live next release. Until then feel free to build locally.

PR #466, opened 2016-08-08
Adds support for Downcast (Mac OS)

gcox:
#355 

~~An update for Downcast which adds the required AppleScript support for this is currently awaiting review in the Mac App Store (submitted on 8/7/16).~~ The most recent version of [Downcast (v2.9.16)](https://itunes.apple.com/us/app/downcast/id668429425?mt=12) contains the required AppleScript support.

### DIFF
--- a/BeardedSpice.xcodeproj/project.pbxproj
+++ b/BeardedSpice.xcodeproj/project.pbxproj
@@ -8,34 +8,159 @@
 
 /* Begin PBXBuildFile section */
 		0301233A1868B1F90080115F /* BeardedSpiceUserDefaults.plist in Resources */ = {isa = PBXBuildFile; fileRef = 030123391868B1F90080115F /* BeardedSpiceUserDefaults.plist */; };
+		0320C0521868C6C2009591CA /* beard.png in Resources */ = {isa = PBXBuildFile; fileRef = 0320C0511868C6C2009591CA /* beard.png */; };
+		0320C0541868C7A4009591CA /* beard-highlighted.png in Resources */ = {isa = PBXBuildFile; fileRef = 0320C0531868C7A4009591CA /* beard-highlighted.png */; };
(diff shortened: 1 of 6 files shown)


gcox:
If a Downcast promo code is needed in order to get this merged, please let me know.

1ps0:
It would help. Send one to my email, it's in the commit logs :)

Stillness-2:
Hello @gcox 
I am trying to develop BeardedSpice on a fork of Beardie :)
https://github.com/Stillness-2/beardie

At the moment, I check and restore adapters and strategies to a working state.
Checking the Downcast adapter found that rewinding to the next/previous track (podcast episode) is not very suitable for a podcast player. I decided to do it like this:
```
- (BOOL)next{
    DowncastApplication *downcast = self.downcast;
    if (downcast) {
        DowncastNowPlayingInfo *npi = downcast.nowPlayingInfo;
        NSInteger newPosition = npi.playPosition + F_SEEK;
        if (newPosition > npi.duration) {
            newPosition = npi.duration;
        }
        [npi setPlayPosition:newPosition];
        return YES;
    }
    return NO;
}
```
Found that it does not work :) On my test, `duration` is always equal ` playPosition`, and setting a new value for `playPosition` does not work.
Can you fixed the `AppleScript` support in` Downcast` for this case? :)

gcox:
@Stillness-2 Thanks for reporting this. There are a couple of issues here. 

Thanks for reporting the issue about the duration & play position always matching, that is definitely a bug and it will be fixed in the next release later this week.

Downcast's now playing info is intentionally readonly. In Downcast's iOS app, there is a setting for controlling what happens when the next/previous remote commands are received. Users can choose to have it skip to the next/previous episode **or** they can have it skip forward/backward by an interval that is configurable within the app. That particular setting doesn't yet exist in the macOS app, but it will in the very near future. The scriptable next/previous commands will honor that setting when it is added. Therefore, there's no need to change Downcast's `NativeAppTabAdapter` implementation.

PR #475, opened 2016-08-24
Fix mike not working after it's unplugged and plugged back in

raja-baz:
I noticed this bug while using the app. If you unplug apple earphones and then plug them back in, media controls from the headphones stop working(volume +- and toggle). Running refreshMikeys a few seconds after headphones are plugged back in seems to fix the issue.


1ps0:
@raja-baz Thanks for doing this, getting this tested and merged.

A few notes that will help in the future.

- Use `git mv` to rename files instead of removing/adding them wholesale. It makes everyone's lives so much easier.
- Keep the core changes and the file renaming as separate commits. It makes my job much easier to see 'oh hey he changed this one little bit' and then 'oh ok this is all just spam about changing a file name'. It's the little things :)

Thanks!

raja-baz:
On mv: agree. On the rename, I agree in general, I debated doing it,
actually. Except in this case part of the logical change was a class name
change... So I wasn't sure which was better.

1ps0:
It makes sense and you did it cleanly, `mv` aside. So thanks for that!

Every time I look at the code base I have to restrain myself from renaming all the things to proper Apple NS Standards. I have a small forest of branches where I started...

PR #487, opened 2016-09-28
Mixcloud beta compatibility

davarisg:
Hey guys,

I am one of Mixcloud's developers and we recently launched our new beta site. You can go [here](https://www.mixcloud.com/beta/join) if you would like to join.

The CSS class names on the player have changed so Bearded Spice is currently broken on the Beta site. I created a second MediaStrategy called `MixcloudBeta.js` since both sites are running at the moment. When we launch the new site I will make the necessary changes and merge the two files.

### DIFF
--- a/BeardedSpice/MediaStrategies/MixCloud.js
+++ b/BeardedSpice/MediaStrategies/MixCloud.js
@@ -10,7 +10,7 @@ BSStrategy = {
   displayName:"MixCloud",
   accepts: {
     method: "predicateOnTab",
-    format:"%K LIKE[c] '*mixcloud.com*'",
+    format:"%K LIKE[c] '*www.mixcloud.com*'",
     args: ["URL"]
   },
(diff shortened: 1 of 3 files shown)


davarisg:
@Stillness-2 when should we expect 2.1.1 to land?

1ps0:
@davarisg this PR has already been merged and made live. Any web-based strategy can be pulled down with Compatibility Updates. 

2.1.1 is around the corner, and not the 'ten years from now' corner. I think we have enough changes that we'll start wrapping it up in the coming weeks.

davarisg:
@1ps0 thats amazing, I wasn't aware of the Compatibility Updates.

1ps0:
@davarisg yep! that was the primary reason for the shift to 2.0. Its easy to run and no big deal if you never know its there (unless you want the new stuff). The name isn't as good or obvious as I'd like though, so if you have suggestions...

davarisg:
@1ps0 how about `Check for pre-release updates`?

PR #503, opened 2016-10-26
Add strategy for be-at.tv

thatmarvin:
The play/pause toggle works, the next button works, and there's no "favorite" feature on be-at.tv. As far as I can tell, the back button doesn't actually do anything on the site, so the keyboard shortcut here doesn't work as well.


1ps0:
All merged and live. Be sure to run compatibility updates.

PR #514, opened 2016-11-23
Added strategy to deal with band camp white label sites

bram2000:
Some record labels are now white labelling Bandcamp, so the URL no longer contains any reference to bandcamp.com. This means that BeardedSpice fails to recognise these as Bandcamp.

I have added a new Media Strategy (copied from bandcamp.js) which detects these white labels by reading the siteroot variable. Tested successfully on http://furtherrecords.org/album/anterior-space and http://recordstore.planetgroucho.com/album/motion-sickness

Thanks!

### DIFF
--- a/BeardedSpice/MediaStrategies/BandCampWhiteLabel.js
+++ b/BeardedSpice/MediaStrategies/BandCampWhiteLabel.js
@@ -3,7 +3,7 @@
 //  BeardedSpice
 //
 //  Copied from BandCamp.js and editted by Jon Bramley on 23/11/2016.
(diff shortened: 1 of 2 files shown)


1ps0:
Would you mind converting your license to be GPLv3? Just change the one line in the header:

> //  Copyright (c) 2015-2016 GPL v3 http://www.gnu.org/licenses/gpl.html

bram2000:
Have converted the license as requested.

1ps0:
All set. Don't forget to run Compatibility Updates to download the official version. Thanks!

Issue #372, opened 2016-04-03
Udemy pushed new ui


donsalari:
With udemy's recently updates, beardedspice is unable to control the video.

danielehrlich:
To get around this (on Mac OSX), you can install the [BetterTouchTool](https://www.boastr.net/). It used to be free, but now it's like $5 (45 day free trial). I created a keyboard shortcut that when I press it in whatever app I'm focused on, it opens up Google Chrome and hits the "space" key, thus pausing or playing Udemy (and it returns the focus back to my original app). Only thing is you have to be on the Udemy Tab in Chrome (although there are more complicated ways around this).

Open up BTT, go to "keyboard", and then make a new shortcut. Look for `Trigger Predefined Action>>>Controlling Other Applications>>>Send Shortcut to Specific App`. From there select Google Chrome and then record your shortcut to call the space key and you should be good.

1ps0:
Looks to be fixed.

PR #515, opened 2016-11-25
Update Udemy MediaStrategy for Late 2016

nelsonjchen:
* Replaced Previous and Next - The "Previous" lecture button is now
    missing and while "Next" lecture could still be implemented, I opted to
    replace the functionality. I made "Next" scrub forward 15 seconds
    and "Previous" scrub backwards 15 seconds as in the GUI.

* Updated the selectors to handle the newer non-iFrame Angular UI.

### DIFF
--- a/BeardedSpice/MediaStrategies/Udemy.js
+++ b/BeardedSpice/MediaStrategies/Udemy.js
@@ -3,29 +3,30 @@
 //  BeardedSpice
 //
 //  Created by Coder-256 on 10/3/15.
-//  Copyright © 2015 BeardedSpice. All rights reserved.
+//  Updated by nelsonjchen on 11/24/16.
+//  Copyright © 2016 BeardedSpice. All rights reserved.
 //
 BSStrategy = {
-  version:1,
+  version:2,
   displayName:"Udemy",
   accepts: {
     method: "predicateOnTab",
     format:"%K LIKE[c] '*udemy.com*/lecture/*'",
     args: ["URL"]
   },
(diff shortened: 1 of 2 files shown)


nelsonjchen:
This might close #372.

1ps0:
You wouldn't happen to have a link to something free I can test on? All the existing links i'm finding are black friday prices ($15).

nelsonjchen:
There's a code that can get the price down to $10, `CNETBF2016`, but even that's not free. It also, like black friday, will expire at the end of the night. Anyway, that's just an FYI. 

As for the free courses, they're not on the front page at the moment. You have to go into the categories on the left menu. At the [IT and software section](https://www.udemy.com/courses/it-and-software/other/), there are a bunch of free courses right above the fold. Here's [one example from there](https://www.udemy.com/aws-concepts/) that can be used. As far as I can see, my approach also work there too. 

If testing on a paid @udemy course is really desired, perhaps they can donate access to one. *shrug*. But as far as I can tell, the viewer on the free courses do not look to be any different from the paid courses.

They do have a `Start Preview` option on a paid course. You can probably use that on something like [this meh course](https://www.udemy.com/refactoru-intro-unit-test/). It does result in an extra element topbar at the top being but it's pretty close to paid.

screen shot 2016-11-25 at 6 16 25 pm

I can say that my approach also appears to work on those as well.

1ps0:
Awesome, best answer ever. Tested and merged. If you dont have it loaded just run Compatibility Updates. Cheers!

PR #522, opened 2016-12-05
Stingray media strategy.

jmcouillard:
Added a media strategy for the canadian web music streaming service.

There is no previous and favorited features in the web version of the app, which explains why they are not implemented.

https://webplayer.stingray.com/

### DIFF
--- a/BeardedSpice/MediaStrategies/Stingray.js
+++ b/BeardedSpice/MediaStrategies/Stingray.js
@@ -1,0 +1,39 @@
+//
+//  Stingray.js
+//  BeardedSpice
+//
+//  Created by Jean-Maxime Couillard on December 5th 2016.
+//  Copyright (c) 2016 GPL v3 http://www.gnu.org/licenses/gpl.html
+//
+BSStrategy = {
+	version: 1,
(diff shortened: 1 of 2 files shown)


1ps0:
I got blocked by the paywall while testing. So for the moment it's up with [2.2.0-b1 prerelease](https://github.com/beardedspice/beardedspice/releases/tag/v2.2.0-b1).

jmcouillard:
Let me know if you want to schedule a remote desktop session so you can validate it (teamviewer or something similar).

1ps0:
Actually if you can just use it normally with the prerelease, it should give me enough data to :+1:

jmcouillard:
Great, just started using it instead of my own build. You should receive the stats (or whatever it is).

1ps0:
It just gives me a simple breakdown of which strategies and their functions are working properly across a slice of OSX versions and browsers that I don't have or use normally :)

1ps0:
@jmcouillard you're good to go. Thanks for your patience!

If you don't have it installed already, run compatibility updates. It pulls down any new strategies :)

PR #539, opened 2016-12-15
Fixed release notes

pdutourgeerling:
This should fix the release notes.
The current https://htmlpreview.github.com/... doesn't work because it redirects to http://htmlpreview.github.io/...

This uses https://htmlpreview.github.io/ so it doesn't redirect to github.io and is served over https


1ps0:
Thank you for this. I've been spending too much brain time trying to figure out what i missed xD

1ps0:
Looks like no cigar. T_T
Is it fixed for you?

PR #545, opened 2016-12-19
Add second Pandora strategy for late 2016 interface update

bretmartin:
Pandora is rolling out a UI update to a subset of its users. Because of this, I think two strategies will need to co-exist for some time. Please let me know if you'd like me to change the naming of this in some way to be clearer.

### DIFF
--- a/BeardedSpice/MediaStrategies/Pandora2016.js
+++ b/BeardedSpice/MediaStrategies/Pandora2016.js
@@ -28,7 +28,7 @@ BSStrategy = {
       'track': document.querySelector('.nowPlayingTopInfo__current__trackName .Marquee__wrapper__content__child').innerText,
       'artist': document.querySelector('.nowPlayingTopInfo__current__artistName').innerText,
       'album': document.querySelector('.nowPlayingTopInfo__current__albumName').innerText,
-      'image': document.querySelector('[data-qa=album_active_image').style['background-image'].slice(5, -2),
+      'image': document.querySelector('[data-qa=album_active_image]').style['background-image'].slice(5, -2),
       'favorited': document.querySelector('[data-qa=thumbs_up_button]').classList.contains('ThumbUpButton--active')
     };
   }
--- a/BeardedSpice/MediaStrategies/Pandora2016.js
+++ b/BeardedSpice/MediaStrategies/Pandora2016.js
@@ -1,0 +1,35 @@
+//
+//  Pandora2016.js
+//  BeardedSpice
+//
+//  Created by Bret Martin on 2016-12-19
+//
+//  Copyright (c) 2015 GPL v3 http://www.gnu.org/licenses/gpl.html
+//
+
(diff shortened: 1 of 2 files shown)


1ps0:
This is a little awkward because you'll have a race condition for which strategy gets applied, and it wont necessarily be the correct one.

Have you noticed any unique identifiers in the url that show up with the redesigned page?

bretmartin:
Thanks for pointing this out. I only tested with this strategy enabled and the other disabled.

The URLs are identical. Is there anything else I can use to differentiate?

bretmartin:
@1ps0 I just double-checked that the URLs and titles are the same between the two interfaces :-(

AnthonyWhitakerJr:
How difficult would it be to combine the strategies? 

For example:
toggle: function () {
    var e = document.querySelector('.playButton');
    var t = document.querySelector('.pauseButton');
    var late2016 =  document.querySelector('.Tuner__Control__Play__Button')

    if(late2016 [insert js version of null-check here]) { late2016.click()}
    else {
        if(t.style.display==='block') { t.click() }
        else { e.click() } 
    }
}

bretmartin:
I didn't even think of that. I'll have a whack at it and test it against both. I'll definitely need a critical eye, though, because this will make the code much more complicated and I'm relatively green with JavaScript.

Should I submit a separate PR or just rebase this one?

1ps0:
If you can get consistent distinction between the two it might be a good short term solution. Go ahead and submit a new PR, cleaner than attempting a rebase.

bretmartin:
Yeah, while the titles and URLs are identical, the document structure is very different.

bretmartin:
@1ps0 @AnthonyWhitakerJr I just noticed the `script` method in the strategy `accepts` element. I'm wondering if using that to test for which UI is in use might be cleaner than a single strategy, since that option will add so much conditional code to every element of the strategy that will be able to be removed in the future when the UI switch is completely rolled out. Untangling all that vs. just removing a strategy -- let me know which approach you'd like me to take.

1ps0:
[`PlexWeb.js`](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/PlexWeb.js) is the only example I think of using the `script` method. Its definitely preferred to the other proposed tangle, but as before it depends if you can find a distinctive trait between the two sites. 

Also as you were saying, since its an A/B test any mess that results of this will be temporary, so it might still be better to keep it to one strategy file.

Also keep in mind you're not constrained to the js functions defined in the strategy. I haven't done a proof of concept, but in theory you should be able to define/invoke functions in addition to those in the template.

bretmartin:
@1ps0 thanks. There are lots of distinctive elements between the two, just not the URL or title.

I didn't realize additional functions were an option -- that would make it a lot cleaner and easier to back out in the future.

Thanks for your thoughts.

PR #546, opened 2016-12-21
BandCamp White Label : be more defensive with siteroot

xavhan:
Sometime the siteroot check could cause errors in unrelated websites, we
should check it before using it.


xavhan:
Do I need to also upgrade the version number ?

1ps0:
Yes please, in the `BandCamp....js` and `versions.plist`.

Holidays are slowing me down a little bit. :)

1ps0:
All set. If you haven't installed it officially, it should update when you run compatibility updates.

PR #550, opened 2017-01-01
Add support for optional new Pandora UI

bretmartin:
Please see https://github.com/beardedspice/beardedspice/pull/545 for some previous discussion on this work.

Because of when things are evaluated (strategy load vs. web page context), it turns out to be pretty tricky to use additional functions as was proposed in my previous pull request. I determined that checking for the presence of `div.Tuner__Controls` reliably detects the new UI, so I used that since it's pretty concise.

I've tested this with both the old and new UIs. I think it makes sense to implement this in the existing strategy so that no one has to modify their BeardedSpice settings: they'll just get support for the new UI automatically, and BeardedSpice will work when they start to use the new UI instead of the old UI. If Pandora drops the old UI, we can simplify this strategy in the future.

### DIFF
--- a/BeardedSpice/MediaStrategies/Pandora.js
+++ b/BeardedSpice/MediaStrategies/Pandora.js
@@ -21,7 +21,7 @@ BSStrategy = {
         document
           .querySelector('.Tuner__Control__Play__Button')
           .attributes['data-qa']
-          .value === 'play_button';
+          .value === 'pause_button';
     } else {
       return document.querySelector('.pauseButton').style.display === 'block';
     }
--- a/BeardedSpice/MediaStrategies/Pandora.js
+++ b/BeardedSpice/MediaStrategies/Pandora.js
@@ -1,10 +1,11 @@
 //
-//  Pandora.plist
+//  Pandora.js
 //  BeardedSpice
 //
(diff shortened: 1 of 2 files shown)


bretmartin:
0a812ae84ca3c2011be1934c3b248e56c1bbbaac has the meat of the new UI support addition. The rest is mostly aesthetic or "administrative". Let me know if you would like anything squashed or separated out into another PR.

1ps0:
@bretmartin In the process of testing this now. Was there an option somewhere to explicitly enable the new UI, or is it a purely random optimizely-style setup?

bretmartin:
@1ps0 You can toggle it, but only as a logged in user (either free or paid, but not as an anonymous "guest"). The toggle is at https://www.pandora.com/account/info in the bottom left and reads either "Go to New Site" or "Go to Classic Site".

1ps0:
@bretmartin Sneaky link hiding at the bottom...
You're all merged and pushed, which means the strategy is live. If you haven't installed it officially, running Compatibility Updates will pull it down.
Happy 2017 :)

bretmartin:
@1ps0 Thanks! Happy New Year.

PR #554, opened 2017-01-02
Add Amazon Alexa media strategy

bretmartin:
The Amazon Alexa web site has a media player to control Alexa-enabled devices. There is no favorite functionality.

### DIFF
--- a/BeardedSpice/MediaStrategies/AmazonAlexa.js
+++ b/BeardedSpice/MediaStrategies/AmazonAlexa.js
@@ -1,0 +1,34 @@
+//
+//  AmazonAlexa.js
+//  BeardedSpice
+//
+//  Created by Bret Martin on 2017-01-02
+//  Copyright (c) 2017 GPL v3 http://www.gnu.org/licenses/gpl.html
+//
+BSStrategy = {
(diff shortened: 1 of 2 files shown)


bretmartin:
Actually, for some of the services it connects to, there is favorite functionality. I didn't notice that before. I'll add that in if I get to it before this is (hopefully) merged, otherwise, I'll add it in a new PR.

1ps0:
@bretmartin For those of us without an Alexa device (despite millions sold!), testing will be the hard part. If you have a test setup I can replicate w/o a device it would help a lot :D

bretmartin:
@1ps0 Do you have or are you willing to open a (free) Amazon.com account? I think that's required. Then you can use https://echosim.io/ to simulate an Alexa device in your browser. Let me know if you need more of a detailed walkthrough.

bretmartin:
@1ps0 Oh, rats, I didn't realize that doesn't support media playback. Let me see if I can find another solution.

1ps0:
I have access to an amazon account to test with, just not a device. I'll take a look at echosim, let me know if you find anything better!

PR #557, opened 2017-01-06
Implemented PhishTracks.com Strategy

febbraro:
Was unsure is the "pause" action was only called if isPlaying is true, or if I needed to check for that.


1ps0:
@febbraro You're all set. Tested and pushed live. If it isn't already installed, you can download it (and all updates) via the compatibility updates option.

:tada:

PR #558, opened 2017-01-06
Support for phish.in

febbraro:
Added support for phish.in


1ps0:
Dude, way to power through these. Merged and live :)

PR #563, opened 2017-01-10
Mixcloud media strategy changes

davarisg:
Hey guys we've just deployed the beta site to everyone so I've made some changes to the media strategies.


1ps0:
A couple things up front, now that I've had a brief moment to look at the PR

1. At the very least there should be an author credited for the work. Tyler did the initial leg work, whoever did the beta site should be credited accordingly.
2. There needs to be a license as well. Default is GPL3 (`//  Copyright (c) 2016-2017 GPL v3 http://www.gnu.org/licenses/gpl.html`)

3. This is my biggest concern: changing the case of the filename. On a majority of macs (at least with default options i'm familiar with), the filesystem is case insensitive, and this is fine. 
However, on a smaller subset, such as my own, the filesystem is very aware of that lowercase `c`, and now I have two copies of the `Mixcloud` strategy stashed and loaded locally. This is undefined behavior, even though it probably is fine too (as the house burns down around us). 
If you have controllable OCD like myself, I'd ask that you not do a file rename for the sake of a capitalization mistake. 

Thanks :)

davarisg:
Yes removing the comments was a mistake. I ll put it back with Tyler's name and the GPL license.

I guess I could keep the original name of the file, it just looks weird to me with a capital 'C'! I thought that `git mv` should handle the file rename properly and only keep one file around.

1ps0:
Well you wrote the new strategy, so you can leave his name but definitely add yours too. 
A good example is the recent strategy update by Bret Martin for [Pandora's beta site](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/Pandora.js).

davarisg:
@1ps0 should be all set now

1ps0:
Thanks for bearing with me. 

Also be aware that updates don't have a destructive function (right now) by design, so this PR only means new releases wont come bundled with the beta site. Since you work with/for mixcloud, if thats a problem (like hits to the beta subdomain or such) let me know, otherwise cheers!

PR #564, opened 2017-01-10
Update README.md

AnthonyWhitakerJr:
Add websites for 4 new media strategies.


1ps0:
Small suggestion for the future: if you could keep like changes (like additions to the supported list) to one commit, it would help in the long run a lot. :)

PR #567, opened 2017-01-14
Add Pluralsight Strategy

alt3red:
Added strategy for https://www.pluralsight.com/


1ps0:
I've been able to test Lynda properly, but Pluralsight is giving me trouble on the signup. My usual tricks are being denied :\

In the mean time, if you want to split this into two PRs, I can get the first strategy up.

Thanks!

alt3red:
Sounds good man. I just created another branch and submitted that one. 

Thank You!

alt3red:
I found out that is you sign up for a free TFS account, you can get 3 months of free pluralsight so if the mood strikes you can get an account and test that way :)

https://www.visualstudio.com/team-services/

PR #589, opened 2017-02-20
Update Synology.js

jhuebsch:
Tested on Audiostation version 6.0.2-3093. 
Security option "Improve security with HTTP Content Security Policy (CSP) header" must be turned OFF for BeardedSpice to control Audiostation. This setting can be found under Control Panel > Security
Audiostation access and control no longer limited to *.synology.me domains. This was accomplished by changing the predicateOnTab to detect page title pattern.
Added trackinfo. Artist isn't a unique field, rather the last item in a " - " separate list. We grab the last item in the list, however if an artist has a " - " in their name this will only return the last part of their name.
Synology Audiostation doesn't allow image to be loaded by non authenticated (non logged in) applications. So album covers aren't supported.

### DIFF
--- a/BeardedSpice/MediaStrategies/Synology.js
+++ b/BeardedSpice/MediaStrategies/Synology.js
@@ -28,12 +28,16 @@ BSStrategy = {
     var albumArtist = document.querySelector('.info-album-artist span').innerText.split(' - ');
     var album = albumArtist[0];
     var artist = albumArtist.reverse();
-    // Image not used due to image resources can only be loaded by authenticated apps
-    // var albumArt = document.querySelector('.player-info-thumb').getAttribute('src');
+    var progress = document.querySelector('.syno-as-player-position-block').innerText;
+    /*
+    Image not used due to image resources can only be loaded by authenticated apps
+    var albumArt = document.querySelector('.player-info-thumb').getAttribute('src');
+    */
     return {
       'track': track,
       'artist': artist[0],
       'album': album,
+      'progress': progress
     };
   }
 }
--- a/BeardedSpice/MediaStrategies/Synology.js
+++ b/BeardedSpice/MediaStrategies/Synology.js
@@ -7,9 +7,10 @@ BSStrategy = {
   version: 2,
   displayName: "Synology Audio Station",
   accepts: {
-    method: "predicateOnTab",
-    format:"%K LIKE[c] '* - Audio Station*'",
-    args: ["title"]
+    method: "script",
+    script: function () {
+        return (baseURL == "webman/3rdparty/AudioStation");
+    }
   },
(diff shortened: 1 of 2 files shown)


1ps0:
Still in the process of setting up to test this. Out of curiosity have you tried setting up a LetsEncrypt (https://letsencrypt.org/getting-started/) ssl certificate instead of just disabling it entirely?

jhuebsch:
I am using a LetsEncrypt certificate, the Security option "Improve security with HTTP Content Security Policy (CSP) header" is different then https. CSP headers (my understanding) help block against XSS attacks. Most of myunderstanding of CSP headers comes from the write up github had  recently about their use of CSP headers https://githubengineering.com/githubs-post-csp-journey/

1ps0:
Interesting, and informative. Thanks for sharing that link.

On a side note though, I find myself wondering, in the PR, about the use of title to identify the page. Is there a way you could use the script option instead? 
There's probably a javascript or html element that is unique to each version that you can use to more reliably identify the Audio Station without potentially grabbing a large part of the random internet along the way :)

jhuebsch:
Script detection would be a much better, I'll see what I can figure out.

jhuebsch:
After testing the last few days, the new detection style, BeardedSpice recognizes the Audiostation tab much quicker. Before with the title detection there was a noticeable delay between pressing the key and the action taking affect.

1ps0:
Forgot to mention, can you update the `versions.plist` as well?

1ps0:
Sorry this is taking me a bit longer than expected to test. I have an older synology device that i need to get updated, in the mean time life is demanding :)
Working on it asap. Thanks for your patience.

jhuebsch:
FYI In the last day of testing I started getting an error on retrieving trackinfo in Chrome 57.0.2987.98 (64-bit) and Firefox 52.0.1 (64-bit). No problems in safari 10.0.3. Of course right before I'm going to not be able to work on this for the next week or so.

1ps0:
That's super helpful, and might extend to other strategies so thanks for pointing that out!

jhuebsch:
It appears that the way Chrome parses the javascript strips all the line breaks. This caused an issue with the single line comment style I was using. I switched to multi line style comment and now trackInfo can be retrieved. 

Added support for the "Show track progress" option.

1ps0:
I got audiostation setup on my nas, just need free time to test it :P

jhuebsch:
FYI I discovered that Synology has a live demo available with their Audiostation package enabled. I was able to start an instance and briefly test Bearded Spice. https://www.synology.com/en-us/dsm/live_demo

This might be an easier test if you don't already use Audiostation and have music stored there.

1ps0:
Yea I was having trouble mounting my music. 

That demo is amazing! Like 1.2.3. done. Pushing this now. Thanks for your help and patience :D

1ps0:
:tada:

PR #590, opened 2017-02-21
Added NPR Strategy, Updated Bandcamp Strategy

egoberdon:
+ Added an NPR Strategy
  + There is a strategy for the NPR One radio website, but not for the NPR website (where they often stream whole albums before they are released), so I added this strategy
  + I left a couple @TODOs in track info where the JS could be nicer, but it all works correctly now
  + I'm not sure if this strategy will conflict with NPR One strategy based on how the url is parsed (since mine is matching on a more general URL) but the NPR One strategy was not working for me, so I was unable to test if there is a conflict

+ Updated BandCamp Media Strategy
  + Simply added an image grabber to the track info portion


1ps0:
You're good to go. Just run Compatibility Updates if needed.

A small note for the future: If you wouldn't mind breaking up a PR into individual strategies/commits, it'd make testing/merging them faster, especially if I get hung up on testing a particular site.

Thanks!

egoberdon:
Sounds good. Noted for the future!

PR #596, opened 2017-03-01
Updated Saavn.js to v2

mohakshah:
- Added trackInfo (name, album, album-art)
- Implemented the isPlaying function
- Tested with Safari 10 and Chrome 56


1ps0:
Tested and pushed to master. You can download it through Compatibility Updates now. 
Thanks for your hard work and patience!

PR #597, opened 2017-03-01
Use "StrategyName - SongTitle" in status menu items instead of the tab title

mohakshah:
The current implementation, which uses a tab's title in the status menu,
works well for sites like soundcloud which update the tab's title according
to the currently playing song. But for sites that _don't_ update the title,
one can't know the current track title simply by looking at beardedspice's
menu. So, instead of using the tab title, we could use the aforementioned
format for the sites whose track info is available and fall back on the
tab title for other sites.

P.S.: The code may be a little amateurish since this is my first time writing in Obj-C. :)


1ps0:
I think this is a good start, if a little heavy handed. We have to consider other sites that might provide the data differently, not have track info, or in general be less friendly to a general approach like this one. 
I'm gonna take a look at how this works in depth with the other websites we support, but in the meantime hold tight. And thanks for the PR :)

PR #600, opened 2017-03-02
ChromeTabAdapter: guard against element removal

wmertens:
While developing, the debugger sometimes fires when I use the media keys, because the added elements are no longer there. A little annoying, this should fix it.


1ps0:
This went live with `2.2.1`. My mistake. Thanks :)

PR #605, opened 2017-03-05
Add GenieMusic

Pnamu:
[http://www.genie.co.kr/](http://www.genie.co.kr/) is music streaming service in South Korea.


1ps0:
@Pnamu Hey! Thanks for making this. 
I only need a few things before I can put this out.

1. Can you `git mv "BeardedSpice/MediaStrategies/Genie Music" BeardedSpice/MediaStrategies/GenieMusic.js`. That will properly name the file so it'll be loaded correctly.

2. Can you add an entry to `BeardedSpice/MediaStrategies/versions.plist` with the strategy name. It would look like:
```
<key>GenieMusic</key>
<integer>1</integer>
```
Please add it in alphabetical order :)

Thanks!

1ps0:
Apologies for the wait on this. Added the `versions.plist` entry myself, but otherwise you're good to go. You can download it through the Compatibility Updates option in the BeardedSpice menu (if you haven't already imported it :)

Thanks!

Issue #421, opened 2016-06-05
Add Egghead.io media strategy


filoxo:
Add a media strategy for [Egghead.io](https://egghead.io/).

---

I've been working on this myself (https://github.com/beardedspice/beardedspice/compare/master...filoxo:eggheadio-mediastrategy). 
- [x] toggle
- [x] previous
- [x] next
- [ ] ~~favorite~~ (no such feature on the website)
- [x] pause
- [x] isPlaying
- [x] trackInfo

I don't have much experience with Mac app development, and can't seem to get it to build in order to test whether its good to merge. The js commands work in a browser console but I'd like to test it in the app before creating a PR. 
When building, the logs show:

```
2016-06-04 22:34:47.600 BeardedSpice[10501:3879921] Failed to load strategy with name: Egghead.io
2016-06-04 22:34:47.604 BeardedSpice[10501:3879921] Failed to load strategy with name: IndieShutffle
```

which is interesting because the strategy is just `Eggheadio` (purposely named to avoid issues) and fixing the typo for `IndieShutffle` (fix [`IndieShutffle` to `IndieShuffle`](https://github.com/beardedspice/beardedspice/compare/master...filoxo:eggheadio-mediastrategy#diff-e756bdabfceefefbf8e3b2b52e394d5aL56)). Not sure what I can do to further chase down why it won't build for me. Any help or insight into how to build would be great. Alternatively, if someone else could build and test that would be great!

PR #617, opened 2017-03-21
Add Egghead.io media strategy

filoxo:
#421 went unattended for a long time. Lots has changed and it looks like its much easier to get stuff working now. This PR adds a media strategy for Egghead.io. I've only tested this by importing the js file to BeardedSpice and everything works correctly.


1ps0:
Excellent. Looks and works great. Thanks for your hard work!

PR #628, opened 2017-04-09
Use Pluralized Strings for Compatibility Updates notification

Mauin:
When checking for compatibility updates, you could get a notification saying `There were 1 compatibility updates.`

This PR adds a `Localizable.stringsdict` file that adds pluralization options for the notification message (including the "No new compatibility updates" message).

I have tested this by manually setting the update count and checking that the notification displays the correct message. 

As I'm by no means an experienced macOS/Obj-C developer, I'm not sure if all the changes in the `project.pbxproj` file are needed and if there are any automated tests for the notification I might have missed. Feel free to point me to anything I might've missed.


1ps0:
Love me a good localization refactor.

PR #634, opened 2017-04-15
Feature/castro

jhuebsch:
Added the Castro.fm media strategy.


1ps0:
@jhuebsch Do you have a test/example link? All I can see is 'get the app'.

jhuebsch:
I only tested against the link provided in the original request ticket. https://castro.fm/episode/PAnZG#5:50

Issue #637, opened 2017-04-20
Synology strategy is causing errors in navigator


xavhan:
https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/Synology.js#L12

BaseURl is not a defined global in chrome. (poke @jhuebsch)
Causing this kind of errors: 

![capture d ecran 2017-04-20 a 13 46 28](https://cloud.githubusercontent.com/assets/1308790/25229230/c9fe1224-25cf-11e7-8c30-cf061e831d42.jpg)

jhuebsch:
That is odd, I'm unable to replicate. I'm able to fully control playback, pause, previous, next with my media keys without any errors appearing in the Chrome console. 

I've tested on my Synology Audio Station (6.0.2-3093) and the[ live demo](https://www.synology.com/en-us/dsm/live_demo) Synology Audio Station (also 6.0.2-3093) with Chrome Version 58.0.3029.81 (64-bit) MacOS.

jhuebsch:
I see the issue. The error occurs on other sites when BeardedSpice checks their baseURL. I'll try and figure out a better check, any suggestions?

xavhan:
I did something similar with bandcamp https://github.com/beardedspice/beardedspice/pull/546

jhuebsch:
@xavhan could you test my pull request https://github.com/beardedspice/beardedspice/pull/638, thanks.

1ps0:
@xavhan @jhuebsch  I totally didnt recognize that this was more than just on the synology page. Breaking all the things. Check the fix now, its live.

xavhan:
@1ps0 thanks, it seems to work.

PR #638, opened 2017-04-20
Improved Synology Audiostation detection.

jhuebsch:
Changed detection script to make sure the baseurl check doesn't cause errors on sites that do not have it set. Fixing https://github.com/beardedspice/beardedspice/issues/637

### DIFF
--- a/BeardedSpice/MediaStrategies/Synology.js
+++ b/BeardedSpice/MediaStrategies/Synology.js
@@ -1,15 +1,19 @@
 //
 //  Synology.plist
 //  BeardedSpice
-//  
+//
 //
 BSStrategy = {
-  version: 2,
+  version: 3,
   displayName: "Synology Audio Station",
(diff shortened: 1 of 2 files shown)


1ps0:
It still fails for me on a demo instance.

jhuebsch:
I'm not experiencing any issues controlling on the live demo site using chrome or safari. The only code I changed was the detection. 

Could you describe your issue more and I'll try and figure out what's going on.

1ps0:
I'm actually thinking this is systemic for me. I was having issues calling up `baseURL`, but I'm also getting the CSP error on just about every page i try to test. This is all taking more time to debug than I can afford.

Can I get some +1s for those who've verified it works for them?

1ps0:
My tests are getting blocked by Chrome 58. After I fix that I'll give this another whirl.

Issue #336, opened 2016-02-11
Begin playing music after wake up


Yeti-or:
I stop music
close mac
mac go to hibernate
after wake up music begin to play
osx 10.10.05
beardedspice/1.2.3

Coder-256:
When your computer goes to sleep, it automatically pauses most processes, including BeardedSpice and your web browser. When it awakes, these processes are resumed. Your music should continue where you left off (unless it is live). What is the problem?

Coder-256:
If you don't want this, you _would_ be able to change this behavior by stopping the music on wake using [SleepWatcher](http://www.bernhard-baehr.de/), [Scenario](http://www.lagentesoft.com/scenario/index.html), [Power Manager](https://www.dssw.co.uk/powermanager/) or similar. However, there must first be an API for BeardedSpice made available. See #332.

Coder-256:
@Yeti-or Did that fix the problem?

Stillness-2:
> I stop music

В чем музыка то играла? браузер/сайт? или там iTunes?

chriswait:
@Coder-256: I think you may have misunderstood the original issue.

> I *stop music*
> close mac

If the music is paused before the mac is closed (and sleeping), we wouldn't expect it to start playing automatically upon waking (i.e this is unexpected behaviour). This seems unrelated to the resuming of processes, and might not be resolved using a B.S API, as you suggest.

Coder-256:
@chriswait Yes, you might be right about the music being stopped before hibernating. However, this would still be fixable with a BS API, as I described. @Yeti-or, it would be helpful to know more information such as logs, browser version, and website.

aray:
I have this problem also with BS 2.2.1 and Pandora Plus. Only happens when BS is running so definitely some sort of bug.

1ps0:
It's part of the automatic handling. I'll see if I can add a workaround.

PR #642, opened 2017-04-29
Pandora: fix pause function in new UI

aray:
Currently with the Pandora new UI the pause function unconditionally clicks the play/pause button. This can lead to odd things like issue #336 and starting playback when headphones are unplugged. This patch checks to see if the button is in fact the pause button, using the same logic as other functions use, before clicking it.


1ps0:
Can you update the `versions.plist` file with the new version number? :)

PR #649, opened 2017-05-19
BBC Radio Programmes Support

bpollman:
Support for BBC's programmes site (http://www.bbc.co.uk/programmes) which is totally different than the live streaming iplayer site (http://bbc.co.uk/radio/player/)

Supports all beardedspice media strategy options and tested with several links on the site.


1ps0:
Hey! Thanks for building this. In the process of (finding time and) testing it, in the meantime can you change the extension of the strategy to `.js`? It should be as simple as `git mv BBCRadioProgrammes.bsstrategy BBCRadioProgrammes.js`

Thanks! I expect to have a lot of time in the next couple days to catch up on the backlog :)

bpollman:
Awesome! Not having media key support on a site after it works perfectly everywhere else was driving me nuts. Thanks for a building a great project.

PR #650, opened 2017-05-19
Adding uninstall.sh script

### LINKED ISSUE #627: How do I uninstall BeardedSpice on mac?
How do I delete all the related files and uninstall BeardedSpice on mac.
Also in System Preferences, Notifications there is and entry (where do I remove this on disk, not possible in the GUI).

jacklightbody:
Should fix #627 

Removes all the associated files and restarts the notification center so that bearded spice gets removed from it. Should probably be updated in the future to a more robust method of finding the app. Wasn't sure if anything should be added to the readme or not, but it can be run with `sudo ./uninstall.sh`

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -8,6 +8,7 @@ sudo rm -rf ~/Library/Application\ Support/BeardedSpice
 sudo rm -rf /Applications/BeardedSpice.app
 
 # Reboot the notifications service so it realizes that the app is gone
-sudo killall usernoted
+launchctl stop com.apple.usernoted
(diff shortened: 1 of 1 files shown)


1ps0:
Instead of `killall usernoted` you might use `launchctl stop com.apple.usernoted; launchctl start com.apple.usernoted`. In my case I didnt need sudo for that, and its less dangerous than a hard kill to a namespace.

jacklightbody:
Good call- just updated

1ps0:
Good to go. Thanks!

PR #658, opened 2017-05-24
Provide Strategy for listen.moe

nVitius:


nVitius:
@1ps0: I've confirmed that this works by importing it into my local installation of BS (v2.2.2) on Chrome 58 and testing it on the site.

nVitius:
I changed the string for the LIKE a little bit. Now, it should match any page on the main app, but not the forums.

PR #659, opened 2017-05-24
Experimental/volume control for native apps

Stillness-2:
**Mac OS X 10.10 or greater required!**

Added the ability to control the sound volume into native apps.

Added the ability to connect applications that do not have support for the Applescript. If they have control points in the menu bar. :)

Added support Tidal Desktop app.
A small step has been taken to the audiophiles who listen to the Tidal through the external DAC.:) Volume control on the DAC, which is connected in exclusive mode, using of the standard keyboard volume buttons.

Also, Preferences pane split on 3 tabs.


BeardedSpice/BSTrack.m
```diff
@@ -18,19 +18,21 @@
 NSString *const kBSTrackNameFavorited = @"favorited";
 NSString *const kBSTrackNameIdentifier = @"BSTrack Notification";
 
+const NSString *kImageLock = @"kImageLock";
```
1ps0: This is an NSString const, which as a class doesn't make sense. We want the pointer to be const and unchangeable. https://stackoverflow.com/a/15157226

BeardedSpice/BSTrack.m
```diff
@@ -18,19 +18,21 @@
...
-@property (nonatomic, strong) NSImage *lastImage;
-
 - (void)_loadImage;
 
 @end
 
 @implementation BSTrack
 
+static NSString *_lastImageUrlString;
+static NSImage *_lastImage;
```
1ps0: Please leave these as instance variables. This makes them singleton values that aren't properly ARC controlled and can have stale or invalid content that crosses over between BSTrack and tab instances.

BeardedSpice/BSVolumeControlProtocol.h
```diff
@@ -0,0 +1,30 @@
...
+//  BeardedSpice
+//
+//  Created by Roman Sokolov on 15.05.17.
+//  Copyright (c) 2015-2017 GPL v3 http://www.gnu.org/licenses/gpl.html
+//
+
+#ifndef BSVolumeControlProtocol_h
+#define BSVolumeControlProtocol_h
+
+typedef enum {
```
1ps0: Please use the objc convention `BSVolumeControl-Protocol.h` for the filename.

Please use the apple helper `typedef NS_ENUM(NSInteger, BSVolumeControlResult) {}` for all enums.
http://nshipster.com/ns_enum-ns_options/

BeardedSpice/runningSBApplication.m
```diff
@@ -117,12 +117,122 @@ - (void)makeKeyFrontmostWindow{
...
+    
+    if (! indexPath.length) {
+        return nil;
+    }
+
+    AXUIElementRef item = nil;
+    AXUIElementRef ref = AXUIElementCreateApplication(self.processIdentifier);
+    
+    if (ref) {
+
```
1ps0: Can we keep the amount of whitespace to a minimum? Especially after a conditional.

BeardedSpice/runningSBApplication.m
```diff
@@ -117,12 +117,122 @@ - (void)makeKeyFrontmostWindow{
...
+    AXUIElementRef ref = AXUIElementCreateApplication(self.processIdentifier);
+    
+    if (ref) {
+        
+        CFIndex count = 0;
+        CFArrayRef items = nil;
+        AXUIElementRef menu = nil;
+        BOOL notFound = NO;
+        if (AXUIElementCopyAttributeValue(ref, (CFStringRef)NSAccessibilityMenuBarAttribute, (CFTypeRef *)&menu) == kAXErrorSuccess
+            && menu) {
```
1ps0: Can we put this into a variable assignment, or otherwise reduce the conditional enough to keep the `if()` on one line?

BeardedSpice/UI/BSVolumeWindowController.m
```diff
@@ -0,0 +1,158 @@
...
+
+#define BASE_FONT_NAME              @"HelveticaNeue-Medium"
+#define BASE_FONT_SIZE              36.0f
+
+@implementation BSVolumeWindowController {
+    NSTimer *_hideTimer;
+}
+
+
+static BSVolumeWindowController *singletonBSVolumeWindowController;
```
1ps0: Unless the var `singletonBSVolumeWindowController` is used elsewhere (which it shouldn't be), can we keep this static variable in the `singleton` method please?

BeardedSpice/UI/BSVolumeWindowController.h
```diff
@@ -0,0 +1,34 @@
+//
+//  BSVolumeWindowController.h
+//  BeardedSpice
+//
+//  Created by Roman Sokolov on 08.05.17.
+//  Copyright © 2017 BeardedSpice. All rights reserved.
+//
+
+#import <Cocoa/Cocoa.h>
```
1ps0: We don't need to import Cocoa. It's all in the Prefix.pch file now.

BeardedSpice/UI/BSVolumeWindowController.h
```diff
@@ -0,0 +1,34 @@
...
+
+typedef enum {
```
1ps0: Another `NS_ENUM` pls.

BeardedSpice/Tabs/TidalTabAdapter.m
```diff
@@ -0,0 +1,593 @@
+//
+//  TidalTabAdapter.m
+//  BeardedSpice
+//
+//  Created by Roman Sokolov on 21.05.17.
+//  Copyright © 2017 BeardedSpice. All rights reserved.
```
Stillness-2: yes

BeardedSpice/AppDelegate.h
```diff
@@ -64,6 +64,8 @@ extern BOOL accessibilityApiEnabled;
 @property (nonatomic, readonly) NSWindowController *preferencesWindowController;
 @property (nonatomic, strong) BSStrategyVersionManager *versionManager;
 
+- (IBAction)TEST:(id)sender;
```
Stillness-2: :)

BeardedSpice/BSTrack.m
Stillness-2: hm :)

BeardedSpice/BSTrack.m
Stillness-2: "This makes them singleton values" - this is what i mean. This should be the case.

BeardedSpice/BSVolumeControlProtocol.h
Stillness-2: About "BSVolumeControl-Protocol.h" - Not important, maybe next time.
NS_ENUM - This is interesting :)

BeardedSpice/runningSBApplication.m
Stillness-2: I put so many blank lines, since it is convenient for me to read the text. I usually put empty line after the beginning of the block of code (after '{'). But it is not important. :)

BeardedSpice/runningSBApplication.m
Stillness-2: no :) It is simple enough.

BeardedSpice/UI/BSVolumeWindowController.m
Stillness-2: I think, yes.

BeardedSpice/UI/BSVolumeWindowController.h
Stillness-2: ok :)

1ps0:
@Stillness-2 Other than the runtime questions I noted in the main review comment, the code looks good.

Stillness-2:
@1ps0 Did you receive my message on mail about the `distr` branch? :)

Stillness-2:
@1ps0 Do you have no objections to raising the minimum version to 10.10?

1ps0:
@Stillness-2 I didnt think you needed a response for the email. Looks like you have it under control.

Did you see my questions about how this PR actually works?

Stillness-2:
@1ps0 what questions do you mean? I didn't find any new questions. I don't understand.

1ps0:
```
Some questions and issues:

- Maybe it's just my test env, but I couldn't remove custom strategies at all. No gear icons appeared at all.
- 'Use current media controller ...' -> checking this also enforces 'Show track progress...' as being enabled/disabled. Is this intentional?
- Other than adding lag to the volume control, what exactly is this feature supposed to do?
```

Stillness-2:
@1ps0 
> Maybe it's just my test env

I Think so.. check this: `Debug-Config.xcconfig`

>'Use current media controller ...'

This is strange bug in IB when you copy `checkbox`. Fixed this in last commit on `master`.

>Other than adding lag to the volume control

Yes, lag is fundamental :( 
I disabled by default this feature in preferences. Also I think about optimisation, but chances are minimal. :( 

>what exactly is this feature supposed to do

Potentially - Added the ability to control the sound volume using keyboard into native apps.
Practically - `TIDAL Desktop app support` use this feature for volume control, when app send sound on not system default device.

raffaelemasotti:
Hi @Stillness-2 ! I tried to download the v.2.2.3, but don't seems to work with TIDAL Desktop (can't see it in applicable tabs)... where I could find a build of the volume-control-for-native-apps branch? :)

Stillness-2:
@raffaelemasotti debug build of the 3.0 version:
https://github.com/beardedspice/beardedspice/issues/771#issuecomment-376469908

"I can see TIDAL in tabs as well, but actually nothing happens to the TIDAL volume." - TIDAL volume indicator is changed from system volume buttons only if sound device for output, in TIDAL preferences, differ from system default and mode of the direct output was selected (audio device locked on TIDAL) .

raffaelemasotti:
@Stillness-2 thank you for your reply. I tried the 3.0 build and I'm using an external dac (selected in TIDAL pref as direct output).. I also checked "use current media controller for the volume control" in bearderdspice preferences, but nothing happens when I use the volume control, as you can see in this record: https://cl.ly/0S0M2c3N0O0q

I noticed that the tab check disappear when I use the volume control. Is there something else that I could try?

Stillness-2:
@raffaelemasotti 
https://youtu.be/wClW6j8_SqQ
Only in "Use Exclusive Mode" TIDAL Menu is used for controlling of the sound volume.
Also we see some switching lags after changing audio settings in TIDAL, it because TIDAL saves preferences in delayed manner.
We have 3 situations:
1. Audio device the same in TIDAL and system output;
2. Audio device differ in TIDAL and system output;
3. Audio device in "Use Exclusive Mode" in TIDAL;

BeardedSpice
1. do nothing. (volume buttons is processed by OS)
2. intercepts volume buttons and changes volume on appropriate device through API;
3. intercepts volome buttons and changes volume through TIDAL Playback Menu. (In this case "mute" do not work, because no such command in TIDAL Menu).

raffaelemasotti:
@Stillness-2 it seems we have the same hardware and configuration. I tried as well to be in case 2 and 3, but BeardedSpice doesn't seems to intercept the volume buttons. Maybe there is something in my system that conflict with that.

Thank you anyway for your help. I'll try to figure out a way to debug it :)

PR #661, opened 2017-05-25
Use latest queries in strategy for HypeMachine

### LINKED ISSUE #657: Hype Machine broken
Hello,

It's been a while since I haven't been on [Hype Machine](http://hypem.com) but BS doesn't seem to map any command anymore.

- macOS 10.12.4
- Google Chrome 58.0
- BS 2.2.2

Thanks,

nVitius:
This fixes #657

### DIFF
--- a/BeardedSpice/MediaStrategies/HypeMachine.js
+++ b/BeardedSpice/MediaStrategies/HypeMachine.js
@@ -3,21 +3,23 @@
 //  BeardedSpice
 //
 //  Created by Jose Falcon on 12/16/13.
+//  Updated to v2 by nVitius on 05/25/17.
 //  Copyright (c) 2013 Tyler Rhodes / Jose Falcon. All rights reserved.
 //
 BSStrategy = {
-  version:1,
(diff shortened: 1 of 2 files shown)


nVitius:
@1ps0: I've confirmed that this works by importing it into my local installation of BS (v2.2.2) on Chrome 58 and testing it on the site.

Stillness-2:
@nVitius What about `trackInfo`?

nVitius:
@Stillness-2 I confirmed that one still works as-is.

1ps0:
@nVitius can you verify that 2.2.3 fixed this issue? I'd rather keep direct js calls than indirect ui element events.

nVitius:
@1ps0 It does indeed work on 2.2.3. That's really odd though, as I remember testing the commands directly in the JS console in Chrome and they didn't work...

1ps0:
It's because Google broke Chrome for a workaround that we normally used, then fixed it in a patch about a week after I released 2.2.2 (about 2-3 months after it broke). The irony is amazing.

PR #683, opened 2017-07-11
[docs] Mention Gaana in README

swaroopch:
It appears that Gaana.com is supported - https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/Gaana.js


1ps0:
Thanks for this!

PR #684, opened 2017-07-14
Add strategy for controlling the web interface of HTTPMS

ironsmile:
[HTTPMS](https://github.com/ironsmile/httpms) is short for "HTTP Media Server". This is a small media server with REST interface I've created.

Not much left to say. As far as I can tell the strategy works well. The only thing not implemented is adding to favorites. This does not make sense for HTTPMS since there is no such a feature. Also, at the moment there is no track/album/artist image support so the key in `trackInfo` is left undefined.

### DIFF
--- a/BeardedSpice/MediaStrategies/httpms.js
+++ b/BeardedSpice/MediaStrategies/httpms.js
@@ -1,0 +1,48 @@
+// Bearded spice strategy for the web interface of HTTPMS.
+// What is bearded spice you ask? https://github.com/beardedspice/beardedspice
+// What about HTTPMS? https://github.com/ironsmile/httpms
+
+BSStrategy = {
+  version: 1,
+  displayName: "HTTPMS",
+  accepts: {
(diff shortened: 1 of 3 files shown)


1ps0:
Hey, thanks for making one of our strategies for your project. 

Since it's such a small project, I might suggest that you instead include your strategy in your HTTPMS repo, with instructions to do custom import. 

Either way, the URL you're using as a predicate is very generic and could conflict with other apps and websites. I might suggest you expand it to include some kind of domain (ex: .com, .net, localhost:8080, etc), or other identifier that will be strongly unique to your project. 

Cheers!

ironsmile:
Hello! Thank you for the fast reply. I would include it in the HTTPMS repo itself. As for the tab matching predicate, title is the only possible solutions since the media server could be installed anywhere and thus have a different URL for everyone of its users. I would guess that most people (myself included) do not use it at localhost.

PR #685, opened 2017-07-15
Update HotNewHipHop

beaumartinez:
Pause/toggle had broken.

Here's a link for testing http://www.hotnewhiphop.com/asap-rocky-liveloveasap-mixtape.31354.html?song-26461

### DIFF
--- a/BeardedSpice/MediaStrategies/HotNewHipHop.js
+++ b/BeardedSpice/MediaStrategies/HotNewHipHop.js
@@ -6,22 +6,22 @@
 //  Copyright © 2015 BeardedSpice. All rights reserved.
 //
 BSStrategy = {
-  version:1,
+  version:2,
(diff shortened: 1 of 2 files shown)


1ps0:
@beaumartinez This one is testing a little wonky for me. The site seems to run on promises, and exceptions out if you attempt to invoke the system while other promise events havent settled. Is that something you can look at more closely?

beaumartinez:
Hey @1ps0, thanks for having time to look at this. What happens exactly? Is it that mashing eg, play/next queues promises? Or does it simply not work for you?

I can have a look in further detail.

1ps0:
It works but if you try to chain actions (play/pause/next/etc) it gets caught in a race condition. I'll have more information later (dont have access to my test machine at the moment). Will update this comment then.

PR #686, opened 2017-07-17
Add favorited track info to GoogleMusic strategy

rodweb:
The selector looks a bit nasty, but it's the only way I figured. To indicate that the track is favorited, Google replaces contents of icon's SVG.

P.S. I'm not feeling confident with setting up XCode and stuff, so any help regarding testing this change will be helpful 🙂 
P.P.S. The former includes throwing at me some guide "for dummies" so I could test it myself.

### DIFF
--- a/BeardedSpice/MediaStrategies/GoogleMusic.js
+++ b/BeardedSpice/MediaStrategies/GoogleMusic.js
@@ -6,7 +6,7 @@
 //  Copyright (c) 2014 Tyler Rhodes / Jose Falcon. All rights reserved.
 //
 BSStrategy = {
-  version:2,
+  version:3,
   displayName:"GoogleMusic",
   accepts: {
     method: "predicateOnTab",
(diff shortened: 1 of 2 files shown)


beaumartinez:
@rodweb If you've got Xcode installed, run `open BeardedSpice.xcworkspace`, and then press the "play" button in the top left, it will compile and run your code (replacing the current instance of BeardedSpice).

rodweb:
Thanks @beaumartinez! I was able to test it out, everything seems to work correctly.

1ps0:
@rodweb @beaumartinez FYI there's an import feature (button below the scrolling site list) that you can use to test out strategies in development.

I've been tight for time recently, but I'll have a look at them asap. Thanks for your hard work :)

1ps0:
Merged and live.  You should be able to fetch the new version with the 'Check for Compatibility Updates' menu option.

PR #692, opened 2017-08-04
Update Audiomack strategy

okcoker:
The Audiomack site had a big redesign. There will be an exposed version of the player for BS in the next deploy.


1ps0:
Hey! Thanks for doing this (and apologies for taking so long to respond).

A couple things real quick while I test.

1. You basically rewrote the whole strategy, so please feel free to add your name to the copyright. It could look something like 
```
//  Created by You on Today's Date.
//  Copyright (c) 2017 GPL v3 http://www.gnu.org/licenses/gpl.html
```

2. Can you update the version number for `Audiomack` in the [versions.plist](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/versions.plist) file?

Thanks!

okcoker:
@1ps0 No worries! Updated things now. I don't think fixing the casing should present any problems right?

1ps0:
One last thing, pretty big but an easy fix. Can you `git mv Audiomack.js AudioMack.js`. Turns out its creating an entirely new strategy (by file namespace) that conflicts with the old one. I can handle it if you just wanna push it out :)

okcoker:
@1ps0 Ok no problem. Just pushed that change

1ps0:
All set. Thanks, and thanks for your patience :)

PR #694, opened 2017-08-09
Update Egghead.io media strategy to v2

filoxo:
Egghead.io's website was updated and the controls no longer worked, this PR fixes that.


1ps0:
All set. Don't forget to run Compatibility Updates. :taco:

PR #697, opened 2017-08-10
BeardedSpice support for beaTunes 5.0.3 or later

hendriks73:
Implemented support for [beaTunes 5.0.3 or later](https://www.beatunes.com/).

### DIFF
--- a/BeardedSpice.xcodeproj/project.pbxproj
+++ b/BeardedSpice.xcodeproj/project.pbxproj
@@ -106,6 +106,7 @@
 		9679334F1855173100A49AC7 /* XCTest.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9679334E1855173100A49AC7 /* XCTest.framework */; };
 		967933501855173100A49AC7 /* Cocoa.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9679332F1855173100A49AC7 /* Cocoa.framework */; };
 		9679336518551BCC00A49AC7 /* ScriptingBridge.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9679336418551BCC00A49AC7 /* ScriptingBridge.framework */; };
+		B19CF39B1F37311500738CBE /* BeaTunesTabAdapter.m in Sources */ = {isa = PBXBuildFile; fileRef = B19CF39A1F37311500738CBE /* BeaTunesTabAdapter.m */; };
 		D09EFF221C8CD49E00B24DA9 /* VLCTabAdapter.m in Sources */ = {isa = PBXBuildFile; fileRef = D09EFF1F1C8CD49E00B24DA9 /* VLCTabAdapter.m */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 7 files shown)


hendriks73:
Anything I can do to help with this?

1ps0:
Native apps tend to be harder to test and take longer to release, because we have to compile and publish a new entire version of the program. I'm working on testing it, but the time between merge and release might be a bit longer than the browser-based strategies.
Appreciate the patience :)

hendriks73:
@1ps0 Please contact me privately, should you need a beaTunes license for testing.

1ps0:
@hendriks73 
Hey! Appreciate the offer and might need to take you up on that as I work through this. 
I had a couple thoughts and questions, but first wanted to verify that you're the (primary) author of this software? So far it looks super useful!

My thoughts:
1. Is there a reason you don't have an AppleScript integration layer? It would be far easier and cleaner to work with, and is the BeardedSpice standard for native applications like this.

2. The Fastsocket library looks to be overkill for the usecase. Correct me if I'm wrong, but the primary use case seems to be Enterprise-grade server-side accessibility for high usage applications, and doesn't really make sense for a simple OS userspace plugin to use when a normal network wrapper provided with Cocoa would suffice. It looks like [NSStream](https://developer.apple.com/documentation/foundation/nsstream) would probably handle the case well. 

Otherwise though, I would really prefer to avoid extra external (library) dependencies when at all possible. We're already loaded up chasing the user-side (per-website, per-native app), adding more internal complexity is something I'd need a very good reason for, and with (1), I think there could be an easier route here. 

Thanks for your patience.

hendriks73:
Thanks, Alex, for looking at this. And yes, I'm the primary author of beaTunes.

1. beaTunes is a cross-platform app (Windows/macOS) and written in Java. AppleScript support for a Java app does not happen easily and is worthless for Windows. Therefore I opted for a platform agnostic solution using simple sockets (connections a-only allowed *from* localhost).
2. Please take look at https://github.com/dreese/FastSocket It's actually a very slim, simple library (two classes, of which I use one). I used it mainly, because it allows me to hide everything network related behind simple interfaces and offers a way to make a *synchronous call*, therefore keeping the rest of the logic similar to what an AppleScript call might look like.

1ps0:
I see the code size of FastSocket. I see the use case as quick short-life TCP transactions. That's fine. The other major concern I have is opening the Application to incoming network requests. I'm in the process of finding a way to determine if it could become a security concern, even if it's only using the loopback interface. 

I'm also looking for alternative ways to support non-applescript compatible applications. I understand the challenges you'd encounter and I want to find a way to work through them with you. I'm not saying add applescript, but there might be other ways to work through it. 

Alternatively I'm debating having a second release path for apps that go outside the current transaction convention (there are several others already). The word bloat has negative connotations, but every extra we add, like FastSocket (there's sqlite now too), incurs a performance cost that affects the entirety of the rest of the app. I can only optimize so far, and especially recently I've had very little time to work directly on the app. 

More on this later. Thanks for your patience.

hendriks73:
> The other major concern I have is opening the Application to incoming network requests.

BeardedSpice would **not** accept incoming network requests. Only beaTunes does. To me that means, that there is no security issue on your side here.

1ps0:
Whoops, bad assumption. Yes, there is no server component to that code. 
I'll pull it in and put it through its paces. Might not be able to merge/push it live for a little while though.
Thanks for working with me here :)

hendriks73:
> Thanks for working with me here :)

No problem at all. Appreciate you're being thorough.

hendriks73:
Just wanted to touch base with you. Any progress? Anything I can help with?

1ps0:
So far just a complete lack of free time. I'm at the end of my vacation now and found that I had almost zero time to work on anything. My daytime is normally consumed by work or work related training, so right now I'm just lining up items to knock down in a row after I've woolgathered a decent amount of time.

Sorry for the delay.

hendriks73:
> So far just a complete lack of free time. I'm at the end of my vacation now and found that I had almost zero time to work on anything. My daytime is normally consumed by work or work related training, so right now I'm just lining up items to knock down in a row after I've woolgathered a decent amount of time.

I understand. Just let me know, when you get to this and you need input.

Thanks!

hendriks73:
Hey... since quite a bit of time has passed... just wanted to check in. Are you guys still interested in merging this eventually?

PR #701, opened 2017-08-18
added siriusxm media strategy

a-chaudhari:
This is for sirius xm's online streaming webapp.

I couldn't find a good image to pluck, but everything else is handled.


a-chaudhari:
Anything I can do to help test this?

1ps0:
Right, sorry for the delay. I think I got stuck at 90% testing it, will wrap that up shortly.

PR #724, opened 2017-10-12
Adds FitRadio media strategy

lucasnad27:
Tested all functionality with my fit radio account

### DIFF
--- a/BeardedSpice/MediaStrategies/FitRadio.js
+++ b/BeardedSpice/MediaStrategies/FitRadio.js
@@ -1,0 +1,39 @@
+//
+//  FitRadio.plist
+//  BeardedSpice
+//
+//  Created by Lucas Culbertson on 10/12/17.
+//
+BSStrategy = {
+  version: 1,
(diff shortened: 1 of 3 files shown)


1ps0:
Hey, thanks for this!

A couple quick things to add while I test this. 

1. A copyright in the header (usually GPL, a-la `//  Copyright (c) 2017 GPL v3 http://www.gnu.org/licenses/gpl.html
`)

2. An entry in [versions.plist](https://github.com/beardedspice/beardedspice/blob/master/BeardedSpice/MediaStrategies/versions.plist) for FitRadio

Cheers!

1ps0:
- [ ] Passes Lint
- [ ] Passes JS Load
- [ ] Passes isPlaying
- [ ] Passes play toggle
- [ ] Passes all other implemented JS

lucasnad27:
I intended the `next` function to allow a user to skip forward to the next track within a mix, which is controlled by `#jplayer_next` div. That's my typical use case for using the next functionality.

PR #730, opened 2017-10-18
Update genie music

kimtree:
Recently, Genie Music changed their player so I changed all source code for supporting new Genie Music player.


1ps0:
Excellent, thanks! Will be testing shortly.

kimtree:
@1ps0 Hi, Thanks for responding. It seems that Genie Music doesn’t blocking access from outside of South Korea. These messages say that the page not found. You can try with the correct url as follow. http://www.genie.co.kr/player/fPlayer

1ps0:
hah, go figure. Works great, thanks for the assist!

PR #742, opened 2017-11-21
Update focus@will strategy

Jayphen:


qayshp:
Nice! Love the way this grabs the track information from the page.
I see that this is using the same object in the DOM as https://github.com/tangledhelix/focusatwill-in-google-chrome/blob/master/alfred_script.js , `faw`. While that makes the most sense, I am seeing this weird bug where `faw.isPlaying` doesn't change from false to true if `faw.play()` is called (and it starts playing), until the second time `faw.play()` is called (and does nothing).
My attempt at toggle used the `playButton` element when found: https://github.com/beardedspice/beardedspice/pull/774/files#diff-a05c76184c0b09914685edb37953437fL16

PR #752, opened 2018-01-08
Added MediaStrategy for 99fm

illBeRoy:
Added MediaStrategy for the Israeli website 99fm (eco99fm.maariv.co.il), which is in the added file `99fm.js`.

Added the strategy to the listings in `versions.plist`.

Since this site serves playlists and not individual tracks, the "next\previous track" buttons actually move the current time forwards and backwards in 30 seconds interval.

Would love a review, thanks :)


illBeRoy:
Ping!
Did I wait long enough for that? ;)

PR #774, opened 2018-04-10
Update for newer FocusAtWill player

qayshp:
Use Substring matching attribute selectors for buttons with class matching play and skip.


qayshp:
Just noticed https://github.com/beardedspice/beardedspice/pull/742 instead. Will probably close this in favor of that, with minor complaint that `faw.isPlaying` seems to be buggy after starting playing with `faw.play()`.

PR #577, opened 2017-02-05
Add Pluralsight strategy

alt3red:
Added strategy for the learning platform https://www.pluralsight.com . I had to recreate this branch because of another strategy that needed removing and was added to it's own branch and merged.

PR #801, opened 2018-06-05
Add strategy for Pluralsight

shreyasminocha:
I added a strategy for [Pluralsight](http://pluralsight.com).


shreyasminocha:
Not sure what happened to #567 and #577.

PR #813, opened 2018-06-24
Added YoutubeMusic Media Strategy

CyrilFind:
It works ok except that the artist does not appear in the menubar dropdown, although it does appear in the notification.


Stillness-2:
@CyrilFind 
```
isPlaying: function () {
         return document.querySelector('paper-icon-button.play-pause-button').attributes['aria-label'] === "Pause"
```
and `pause` method does not work if your system language is not English (at least on Safari)

If you a Russian- `Pause` -> `Приостановить`  :)

PR #856, opened 2018-11-18
fix: pause() in MediaStrategies for Yandex Services

mitinarseny:
On closing the lid of laptop and some other events `pause()` method is called. So if it only toggles play/pause and if I have already paused the track, when I'll close the lid, it will resume playing. The fix is to check if current song is playing at the moment, and then stop it. This is the only way to do it since Yandex's `externalAPI` does not provide separate methods like `play()` and `pause()`.


mitinarseny:
@Stillness-2 Could you review, please?

mitinarseny:
@1ps0 could you kindly take a look on solution?

Issue #829, opened 2018-09-03
BBC iPlayer Update breaks Beardedspice support


bpollman:
BBC's latest update breaks support and will probably limit Next/Prev controls. 

To be honest it looks terrible and I hope they roll it back. 

PS: I'm the author of the existing iPlayer support script. I'll try to get it updated sooner than later.

iliakarasin:
Same issue.

mark-reason:
Any support for the new BBC sounds player?

bpollman:
Finally added support for BBC Sounds in this PR:

https://github.com/beardedspice/beardedspice/pull/884

bpollman:
Any idea if this project is still active? It seems pretty quiet?

Issue #871, opened 2019-01-02
[App Support] BBC Sounds


kassim:
e.g. https://www.bbc.co.uk/sounds/play/m0001sdm

BBC have a new music website which will replace iPlayer radio

bpollman:
Finally added support for BBC Sounds in this PR:

https://github.com/beardedspice/beardedspice/pull/884

PR #884, opened 2019-02-12
BBC Media Strategy Updated to work with BBCSounds

bpollman:
BBC Radio and BBC Radio Programmes media strategies are broken and obsolete since BBC moved to its new BBC Sounds website design

Resolves Issues: #829  #871


Stillness-2:

Issue #219, opened 2015-08-18
[App Support] Player.FM


enjoyfebruary:
Would love to add Player.FM support. 

https://player.fm

PR #901, opened 2019-05-01
Added Player FM support

lchojnowski:
Solves #219 (Player FM)


Stillness-2:

PR #910, opened 2019-06-14
Add support for Amazon Music macOS application

insanoid:
- Since there is no apparent ScriptingBridge integration the only way to interact with Apple Music app is through the menu bar items.
- The information about the current playing track is not accessible anywhere, so it cannot be included in the UI.
- The Amazon Music app is only available in English (no translation files in the app bundle) so the string `Pause` is used to compare without translations.


t-ricci-enhancers:
Using Version 2.2.3 (29) I see it in the lists but is not working with Amazon @insanoid

t-ricci-enhancers:
Is it web only? I am using the desktop app

Issue #578, opened 2017-02-06
[App Support] NTS Radio


marcw:
Hi! It would be super cool to have support for NTS Radio. http://www.nts.live/

Thanks! :)

neilkimmett:
I took a look at adding support for this, figured it'd be easy given they just use a Mixcloud miniplayer (https://www.nts.live/shows/moxie/episodes/moxie-6th-june-2018) but couldn't figure it out. Maybe because the player is in an `iframe` ? Will try and find time to take another look, I listen to NTS a lot so would find this super useful.

ondrejrohon:
I listen to NTS a lot as well, it would be sufficient for me if Beardedspice would work on mixcloud website, because you can listen to everything from NTS there. Unfortunately it's not working with mixcloud...

PR #943, opened 2020-07-08
Support NTS Radio

leveneg:
Add basic functionality for playing & pausing audio from NTS radio
(https://nts.live).

### DIFF
--- a/BeardedSpice/MediaStrategies/NTSRadio.js
+++ b/BeardedSpice/MediaStrategies/NTSRadio.js
@@ -1,0 +1,52 @@
+//
+//  NTS Radio
+//  BeardedSpice
+//
+//
+//  Created by Grant Levene on 07/08/20
+//  Copyright (c) 2015-2019 GPL v3 http://www.gnu.org/licenses/gpl.html
+//
+
+BSStrategy = {
+    version: 1,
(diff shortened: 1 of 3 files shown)


leveneg:
This should help address #578 .

Couple other things to note (that I should have just included in the commit message):

- When listening to audio from NTS.live, it's from one of two sources:
    1. An `<audio>` element
        - easy enough- just `querySelector` for it and `play` / `pause` it.
    2. An `<iframe>` with a Mixcloud widget inside.
        - Bit trickier. While mixcloud widgets do have a decent [API](https://www.mixcloud.com/developers/widget/), it goes through [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) asynchronously, which makes returning data difficult as BS cannot handle promises, near as I can tell (though I would be delighted to be wrong).

Stillness-2:
@leveneg You may convert PR for that repo https://github.com/beardedspice/BS-Strategies, or simple export `.bsstrategy` from you BreadedSpice app. I will merge PR in BS-Strategies with minimal checks.

leveneg:
@Stillness-2 [sure thing](https://github.com/beardedspice/BS-Strategies/pull/2)

PR #973, opened 2021-02-16
Add support for Quod Libet.

### LINKED ISSUE #971: FR: Support Quod Libet
[Quod Libet](https://quodlibet.readthedocs.io/en/latest/) is a cross-platform audio player, and I _really_ miss being able to to control playback with media keys since upgrading to MacOS Catalina.

Could BeardedSpice add support for this?

You can trivially control the app [via a FIFO pipe](https://quodlibet.readthedocs.io/en/latest/guide/interacting.html#controlling-the-player), just write out [one of the supported messages](https://github.com/quodlibet/quodlibet/blob/b92e08484e44f7e056a3fb3f7daeb70b201698c9/quodlibet/cli.py#L76-L82) (encoded to UTF-8 bytes) to the pipe. For the functions that BeardedSpice supports, only the `next`, `previous`, and `play-pause` commands would be needed. A command ends with a newline (don't forget to flush the buffer if you opened the FIFO with anything other than unbuffered or a line buffer).

You can also read the current playing track, from `~/.quodlibet/current`, which contains the full set of [track tags](https://quodlibet.readthedocs.io/en/latest/guide/tags/tags.html), newline-separated, encoded to UTF-8. It should be trivial to extract the `title` and `artist` fields; each line starts with `fieldname=` (where field names _can_ start with `~` or `~#` but otherwise consist of lower-case ascii letters and underscores).

Basically:

- Quod Libet is active if `~/.quodlibet/control` exists and is a FIFO pipe (`S_ISFIFO(path)` is true).
- You can control playback by writing UTF-8 encoded commands to the FIFO pipe, terminated by a newline.
- You can get data on the currently playing song by reading `~/.quodlibet/current`.

My Objective-C fu is practically non-existing, so I am not sure if I could create a competent pull request for this myself.

mjpieters:
Quod Libet is a cross-platform GTK+-based audio player. While it doesn't support Apple scripting definitions, it _does_ expose a named-pipe based query and control API.

This tab adapter makes use of that API to query the player current status and send commands to it.

This closes #971.

### DIFF
--- a/BeardedSpice/Tabs/QuodLibetTabAdapter.m
+++ b/BeardedSpice/Tabs/QuodLibetTabAdapter.m
@@ -151,7 +151,7 @@ static NSString * FormatNSTimeInterval(NSTimeInterval interval) {
         }
         
         _track = [BSTrack new];
-        NSString *basename;
+        NSString *filename;
 
         NSArray *lines = [current componentsSeparatedByCharactersInSet:[NSCharacterSet newlineCharacterSet]];
         for (NSString *line in lines) {
@@ -167,10 +167,7 @@ static NSString * FormatNSTimeInterval(NSTimeInterval interval) {
                 } else if ([key isEqualToString:QL_TRACK_ALBUM]) {
                     _track.album = value;
                 } else if ([key isEqualToString:QL_TRACK_FILENAME]) {
-                    basename = [value lastPathComponent];
-                    if ([NSString isNullOrEmpty:basename]) {
-                        basename = value;
-                    }
+                    filename = value;
                 } else if ([key isEqualToString:QL_TRACK_LENGTH]) {
                     NSTimeInterval length = [value doubleValue];
                     NSTimeInterval position = length * _progress;
@@ -183,10 +180,14 @@ static NSString * FormatNSTimeInterval(NSTimeInterval interval) {
         
         if ([NSString isNullOrEmpty:_track.track]) {
             // No title field for this track, fall back to basename.
-            // QuodLibet's "current" file is guaranteed to have a ~filename tag, so
-            // we can rely on `basename` being set. The format here, using the base
+            // QuodLibet's "current" file is guaranteed to have a ~filename tag, we can
+            // rely on that here. The output format here, using the base
             // filename with localised `[Unknown]` postfix, echoes how QL generates
             // a track title for such tracks.
+            NSString *basename = [filename lastPathComponent];
+            if ([NSString isNullOrEmpty:basename]) {
+                basename = filename;
+            }
             NSString *unknown = NSLocalizedString(@"Unknown", @"QuodLibetTabAdapter");
             _track.track = [NSString stringWithFormat:@"%@ [%@]", basename, unknown];
         }
--- a/BeardedSpice/Tabs/QuodLibetTabAdapter.m
+++ b/BeardedSpice/Tabs/QuodLibetTabAdapter.m
@@ -1,0 +1,364 @@
+//
+//  QuodLibetTabAdapter.m
+//  BeardedSpice
+//
+//  Created by Martijn Pieters on 13/02/2021.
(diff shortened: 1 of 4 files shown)


mjpieters:
This is my first ever foray into Objective-C and Cocoa programming, so I almost certainly made missteps somewhere. Please do review with an eye for those and let me know what needs improving!

BeardedSpice/Tabs/QuodLibetTabAdapter.m
```diff
@@ -0,0 +1,364 @@
...
+
+        NSArray *lines = [current componentsSeparatedByCharactersInSet:[NSCharacterSet newlineCharacterSet]];
+        for (NSString *line in lines) {
+            NSRange delim = [line rangeOfString:@"="];
+            if (delim.location != NSNotFound) {
+                NSString *key = [line substringToIndex:delim.location];
+                NSString *value = [line substringFromIndex:delim.location + 1];
+                
+                if ([key isEqualToString:QL_TRACK_TITLE]) {
+                    _track.track = value;
```
Stillness-2, line 162: Everything's OK! But there are tracks with no metadata. What if to make the file name without the extension as the title of the track in the case when the track has no title.

Stillness-2 (changes requested):

BeardedSpice/Tabs/QuodLibetTabAdapter.m
mjpieters: Fair point; Quod Libet, in that case, uses the filename too with `[Unknown]` added. I'll do the same here.

Stillness-2 (approved):

BeardedSpice/Tabs/QuodLibetTabAdapter.m
```diff
@@ -0,0 +1,382 @@
...
+- (BOOL)isPlaying{
+    [self loadQuodLibetState];
+    return _status == QuodLibetStatusPlaying;
+}
+
+#pragma mark QuodLibet remote control methods
+
+- (void)loadQuodLibetState{
+    if (_status != QuodLibetStatusUnknown) {
+        return;
```
Stillness-2, line 209: @mjpieters Is it legal at all? :)

Stillness-2: there is a feeling that the method will work completely only once...

mjpieters: BeardedSpice creates new instances of each native app tab adapter for every interaction. `_status` is caching the QL status only for the lifetime of each instance.

So, all this works fine for me, I've been using this code for a few days now without issue. :-)

BeardedSpice/Tabs/QuodLibetTabAdapter.m
```diff
@@ -0,0 +1,382 @@
...
+}
+
+- (void)previous{
+    [self sendControlToQuodLibet:QL_CONTROL_PREVIOUS];
+}
+
+- (BSTrack *)trackInfo{
+    [self loadQuodLibetState];
+    if (_status == QuodLibetStatusPlaying) {
+        if (_track) {
```
Stillness-2, line 140: @mjpieters This optimization will lead to the fact that the track data does not change even when playing another track, it seems to me. We need to receive up-to-date track information. :)

BeardedSpice/Tabs/QuodLibetTabAdapter.m
Stillness-2: "BeardedSpice creates new instances of each native app tab adapter for every interaction. " Hm. Sorry, I forgot that is different here :)

mjpieters: I initially had more cache invalidation logic with a timestamp but found that the cache was gone every time, until I realised that the lifetime of any adapter instance is really short.

Per interaction we do want to cache this data as the QL named pipe API is relatively slow.

